PR #3, opened 2016-07-22
Add a "status" endpoint for health checks

jkodumal:


ld-relay.go
```diff
@@ -45,10 +45,13 @@ type Config struct {
 	Environment map[string]*EnvConfig
 }
 
+type StatusEntry struct {
+	Status string `json:"status"`
+}
+
 var configFile string
 
 func main() {
```
jkodumal: I got rid of the wait group, which allows the server to start before all of the environments have been initialized. This makes the whole system a little more resilient-- it can begin serving requests if (e.g.) one environment can't connect.

ld-relay.go
```diff
@@ -74,17 +77,23 @@ func main() {
 
 	handlers := cmap.New()
 
-	wg.Add(len(c.Environment))
+	clients := cmap.New()
+
+	for envName, _ := range c.Environment {
```
jkodumal: This pre-seeds the map with all the configured environments.

apucacao:
lgtm

ld-relay.go
```diff
@@ -101,7 +110,39 @@ func main() {
 		}(envName, *envConfig)
 	}
 
-	wg.Wait()
+	http.Handle("/status", http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
```
drichelson: Do we want to make the port configurable? i assume it will be listening on port 80 by default..

jkodumal: It is-- this isn't where that's handled though. See where the configuration port is usdd

drichelson:
👍

ld-relay.go
pkaeding: What will happen if a connection comes in for an environment that is not yet initialized?

pkaeding:
lgtm
:+1:
looks good to me
merge away
I approve
canned replies are great.

ld-relay.go
jkodumal: The corresponding client won't yet be in the handler map, so you'll get a 404.

PR #4, opened 2016-08-02
Make LD relay optionally behave like LD daemon

jkodumal:
With this change, LD relay subsumes the functionality in LD daemon. You can run it exactly like LD daemon, in which case you do not need to front it with a load balancer. You can also persist flags in Redis, which is independently useful.


drichelson:
Code changes look good. What's the plan for migrating git history/deprecating the ldd repo? We don't want to maintain both.

jkodumal:
Still TBD on the git history...

PR #8, opened 2016-09-08
Adds Dockerfile and a script for building Docker image

ahume:
A somewhat speculative PR. I'm inexperienced with the go language/run-time, so I'm not 100% sure if this is good. But we're running the relay in Kubernetes, and this is how we're building the image. Even if there's nothing to merge here, would be interested in your feedback from the go context.

The image is here as well: https://hub.docker.com/r/andyhume/ld-relay/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,10 @@
+FROM alpine
+
+RUN apk --update upgrade && \
+    apk add curl ca-certificates && \
+    update-ca-certificates && \
+    rm -rf /var/cache/apk/*
+
+ADD ldr /
+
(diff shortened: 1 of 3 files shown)


ahume:
Of course, we may have missed that there are existing Docker images out there (or you have no interest in supporting a Docker use-case), in which case feel free to ignore/close. :)

jkodumal:
@ahume thanks for the contribution-- @pkaeding has a little feedback but we're definitely going to support and publish our own Docker images. This looks like a good start.

jkodumal:
We've now got a Docker build on master based on a different approach. Closing this one out.

PR #9, opened 2016-09-13
(Work in progress) Add support for Redis sentinel

jkodumal:
Also adds support for database and password options for both sentinel and Redis


jkodumal:
We've stopped work on this temporarily. When we move forward, much of this work will probably be in the go-client itself.

bschaeffer:
Is sentinel a supported but undocumented feature in the relay?

PR #11, opened 2016-10-21
Allow the in-memory cache TTL to be specified as a configuration parameter

jkodumal:


README.md
```diff
@@ -72,6 +73,8 @@ To set up LDR in this mode, provide a redis host and port, and supply a Redis ke
         prefix = "ld:spree:test"
         apiKey = "SPREE_TEST_API_KEY"
 
+You can also configure an in-memory cache for the relay to use so that connections do not always hit redis. To do this, set the `ttl` parameter in your `redis` configuration section to a number (in milliseconds).
```
drichelson: I think we should have the in-memory cache's default TTL be > 0. Anyone running LD-Relay for scale reasons will likely want it enabled.

jkodumal: Good idea-- addressed in latest commit.

ld-relay.go
```diff
@@ -47,6 +49,7 @@ type Config struct {
 	Redis struct {
 		Host string
 		Port int
+		Ttl  *int
```
jkodumal: Mmm... true. Suggestions?

drichelson: but if it is consistent with how it is named in the RedisFeatureStore we should either leave it or change it in both places

drichelson: InMemoryTtl?

drichelson:
closing, as this is not quite the right approach

drichelson:
Sorry. I accidentally closed this.

PR #22, opened 2017-11-28
add minwinsvc to enable running as a windows service

kpankonen:
This MR adds [github.com/kardianos/minwinsvc](https://github.com/kardianos/minwinsvc) to make it possible to run ld-relay as a Windows service. There is no effect on non-Windows builds.


ashanbrown:
@kpankonen This is merged.  Thanks again.

PR #32, opened 2018-06-14
Added in a way to modify the relay.

dM-At-Bentley:
Using the userDefinedEndpoints EndpointHandler array, a user can create endpoints that do not conflict with the API to do things on their own relay such as allow a traffic manager to ping it. It is advised to consult the LaunchDarkly API to create non-conflicting endpoints. The default is to not have any user defined endpoints. This allows more modification for the user without having to read large chunks of the relay code. This also avoids breaking any relays that will be updated from this.

### DIFF
--- a/ld-relay.go
+++ b/ld-relay.go
@@ -47,6 +47,8 @@ var (
 	configFile        string
 )
 
+var userDefinedEndpoints []EndpointHandler = nil; //update this to use endpoints that are not utilized by the api to add extra features to a relay
+
 type EnvConfig struct {
 	SdkKey        string
 	ApiKey        string // deprecated, equivalent to SdkKey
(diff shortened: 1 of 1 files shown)


ashanbrown:
@dM-At-Bentley Thanks for the suggestion.  This is definitely an interesting application.  Our preference would be to extract the relay into a package that you could import in your relay rather than try to make our "main" package modification-friendly, and then create a separate package that implements the executable.  At that point, it might still be useful to user-defined functions if they wanted access to the per environment context, such as the LaunchDarkly client itself .  On the other hand, I could also imagine an arrangement where the relay is just one of a number of handlers in your application, and you'd connect up your own endpoints separately in your own router.  I suspect that might be sufficient for most users needs.  Please let me know if that sounds like a reasonable alternative.

dM-At-Bentley:
@ashanbrown That does sound like a reasonable alternative.

ashanbrown:
I'm closing this because our latest relay code should now be importable into your own server (as of release 5.1.0).  If simply using `relay` as a http handler isn't sufficient for your needs, please feel welcome to submit a PR that provides access to other methods/objects and we'll be happy to consider it.  Thank you again for your original suggestion.

PR #33, opened 2018-06-25
gopkg.in/launchdarkly/ld-relay.v5 release

ashanbrown:
This package is now known as `gopkg.in/launchdarkly/ld-relay.v5`.

### Changes
* `relay` is available as an http handler for use in other servers.
* Add godoc comments for public methods.
* Enable multiple linters and fix lint issues.
* Switch to multi-stage docker build.


eli-darkly:
Looks like there is a build error, otherwise LGTM.

eli-darkly (approved):

PR #45, opened 2018-10-15
Run docker image as non-root user

sdif:
Hi,

We tried to implement ld-relay on our Kubernetes infrastructure but it runs as root. 
It worked well with this configuration and we thought about sharing it with you as it's a best practices in most companies to run containers as unprivileged users.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,6 +20,11 @@ RUN go build -a -o ldr ./cmd/ld-relay
 
 FROM alpine:3.7
 
+RUN addgroup -g 1000 -S ldr-user && \
+    adduser -u 1000 -S ldr-user -G ldr-user && \
+    mkdir /ldr && \
+    chown 1000:1000 /ldr
+
 RUN apk add --no-cache \
     curl \
     ca-certificates \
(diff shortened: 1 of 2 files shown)


ashanbrown:
@sdif Thanks for the PR.  Could you tell us a bit more about this change and ideally point us to the source of your code for constitutes the proper way of configuring groups and users?  We'd just like to know more about why this is a best practice -- it does seem reasonable --  and also establish that this won't cause problems for users who are not using kubernetes.  Thanks.

ashanbrown:
@sdif This seems to correspond to advice given in https://medium.com/@mccode/processes-in-containers-should-not-run-as-root-2feae3f0df3b.  There is apparently a caveat that you can still get root access on the host from a container in some situations as described https://www.electricmonk.nl/log/2017/09/30/root-your-docker-host-in-10-seconds-for-fun-and-profit/ unless you configure your docker host to remap uids (see https://docs.docker.com/engine/security/userns-remap/).  I think that is beyond the scope of what needs to happen in the container though.  Thanks for the PR.

sdif:
@ashanbrown Hi, thanks for the quick review and sorry for the delay.
I think this is a first step and it will allow users running LDRelay on Kubernetes to run the container without tweaking their settings

missourian55:
Thank you sdiff. We have the exact same requirement. Running the ld-relay in Kubernetes infrastructure as a non root user and +1 for exclusive location of /ldr/ld-relay.conf this should help me run the configuration from k8 config map instead of docker environment variable. 

```CMD ["/usr/bin/ldr", "--config", "/ldr/ld-relay.conf"]```

PR #48, opened 2018-11-12
Update docker-entrypoint.sh

lukasmrtvy:
Support for REDIS_URL environment variable in docker-entrypoint.sh

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -16,21 +16,27 @@ heartbeatIntervalSecs = ${HEARTBEAT_INTERVAL:-15}
 " > /ldr/ld-relay.conf
 
 if [ "$USE_REDIS" = 1 ]; then
+if [ -z "${REDIS_HOST}" ] && [ -z "${REDIS_URL}" ]; then echo "Choose REDIS_HOST and REDIS_PORT or REDIS_URL"; exit 1; fi
 if echo "$REDIS_PORT" | grep -q 'tcp://'; then
-  # REDIS_PORT gets set to tcp://$docker_ip:6379 when linking to a redis container
-  # default to using those values if they exist
-  REDIS_HOST_PART="${REDIS_PORT%:*}"
-  REDIS_HOST="${REDIS_HOST_PART##*/}"
-  REDIS_PORT="${REDIS_PORT##*:}"
-fi
-
+# REDIS_PORT gets set to tcp://$docker_ip:6379 when linking to a redis container
+# default to using those values if they exist
+REDIS_HOST_PART="${REDIS_PORT%:*}"
+REDIS_HOST="${REDIS_HOST_PART##*/}"
+REDIS_PORT="${REDIS_PORT##*:}"
 echo "
 [redis]
 host = \"${REDIS_HOST:-redis}\"
 port = ${REDIS_PORT:-6379}
-localTtl = ${REDIS_TTL:-30000}
+" >> /ldr/ld-relay.conf
+elif [ -n "${REDIS_URL+x}" ]; then
+echo "
+[redis]
+url = \"${REDIS_URL:-redis://:password@redis:6380/0}\"
 " >> /ldr/ld-relay.conf
 fi
+echo "localTtl = ${REDIS_TTL:-30000}" >> /ldr/ld-relay.conf
+fi
+
 
 if [ "$USE_EVENTS" = 1 ]; then
 echo "
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -16,27 +16,29 @@ heartbeatIntervalSecs = ${HEARTBEAT_INTERVAL:-15}
 " > /ldr/ld-relay.conf
 
 if [ "$USE_REDIS" = 1 ]; then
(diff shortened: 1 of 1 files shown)


lukasmrtvy:
There is also problem with SSL, if you want to access REDIS with SSL enabled, it does not work.

lukasmrtvy:
There is also problem with SSL, if you want to access REDIS with SSL enabled, it does not work.

EDIT:
Proper SSLredis URL according IANA is:
rediss://[:password]@[host]:[port]/[db] -> its working

docker-entrypoint.sh
```diff
@@ -16,22 +16,28 @@ heartbeatIntervalSecs = ${HEARTBEAT_INTERVAL:-15}
...
 
 if [ "$USE_REDIS" = 1 ]; then
+if [ -z "${REDIS_HOST}" ] && [ -z "${REDIS_URL}" ]; then echo "Choose REDIS_HOST and REDIS_PORT or REDIS_URL"; exit 1; fi
 if echo "$REDIS_PORT" | grep -q 'tcp://'; then
-  # REDIS_PORT gets set to tcp://$docker_ip:6379 when linking to a redis container
-  # default to using those values if they exist
-  REDIS_HOST_PART="${REDIS_PORT%:*}"
-  REDIS_HOST="${REDIS_HOST_PART##*/}"
-  REDIS_PORT="${REDIS_PORT##*:}"
-fi
```
eli-darkly, line 25: This `fi` shouldn't have been deleted— it goes with the `if` on line 20. Also, the lines inside this `if` block shouldn't have been unindented.

lukasmrtvy: I dont know what You mean... 
Is:
```
if [ "$USE_REDIS" = 1 ]; then
        if [ -z "${REDIS_HOST}" ] && [ -z "${REDIS_URL}" ]; then echo "Choose REDIS_HOST and REDIS_PORT or REDIS_URL"; exit 1; fi

        if echo "$REDIS_PORT" | grep -q 'tcp://'; then
                # REDIS_PORT gets set to tcp://$docker_ip:6379 when linking to a redis container
                # default to using those values if they exist
                REDIS_HOST_PART="${REDIS_PORT%:*}"
                REDIS_HOST="${REDIS_HOST_PART##*/}"
                REDIS_PORT="${REDIS_PORT##*:}"
                cat <<-EOF >> /ldr/ld-relay.conf
                [redis]
                host = "${REDIS_HOST:-redis}"
                port = "${REDIS_PORT:-6379}"
                EOF
        elif [ -n "${REDIS_URL+x}" ]; then
                cat <<-EOF >> /ldr/ld-relay.conf
                [redis]
                url = "${REDIS_URL:-rediss://:password@redis:6380/0}"
                EOF
        fi
        echo "localTtl = ${REDIS_TTL:-30000}" >> /ldr/ld-relay.conf
fi
```
more suitable for You?

eli-darkly: No, there's still a problem there. The problem is that you have made the entire section from `# REDIS_PORT gets set to tcp://$docker_ip:6379` through `port = "${REDIS_PORT:-6379}" \n EOF` conditional on `REDIS_PORT` matching that special Docker URL pattern. That means that if `REDIS_HOST` is set, and `REDIS_PORT` is just a port number, it will _not_ put the host and port in the config file.

The logic should be:

* Write `[redis]` to the config file no matter what (I mean, if `USE_REDIS` is set).
* If `REDIS_HOST` or `REDIS_PORT` is set to _anything_, then we are going to write `host=` and `port=` to the config file. The default values for those are `REDIS_HOST` and `REDIS_PORT`; however, if `REDIS_PORT` starts with `tcp://` then we will parse it into a host and port.

So I think that would end up like this:

```
if [ "$USE_REDIS" = 1 ]; then
        if [ -z "${REDIS_HOST}" ] && [ -z "${REDIS_URL}" ]; then echo "Choose REDIS_HOST and REDIS_PORT or REDIS_URL"; exit 1; fi

        echo "[redis]" >> /ldr/ld-relay.conf

        if [ -n "$REDIS_HOST" || -n "$REDIS_PORT" ]; then
                if echo "$REDIS_PORT" | grep -q 'tcp://'; then
                        # REDIS_PORT gets set to tcp://$docker_ip:6379 when linking to a redis container
                        # default to using those values if they exist
                        REDIS_HOST_PART="${REDIS_PORT%:*}"
                        REDIS_HOST="${REDIS_HOST_PART##*/}"
                        REDIS_PORT="${REDIS_PORT##*:}"
                fi
                cat <<-EOF >> /ldr/ld-relay.conf
                host = "${REDIS_HOST:-redis}"
                port = "${REDIS_PORT:-6379}"
                EOF
        elif [ -n "${REDIS_URL+x}" ]; then
                cat <<-EOF >> /ldr/ld-relay.conf
                url = "${REDIS_URL}"
                EOF
        fi
        echo "localTtl = ${REDIS_TTL:-30000}" >> /ldr/ld-relay.conf
fi
```

lukasmrtvy: You are right. I added a fix for it.

eli-darkly (approved):

eli-darkly:
This will be in the next release. Thanks!

PR #49, opened 2018-11-20
prepare 5.4.4 release

eli-darkly:
### Fixed
- Fixed a bug that would cause the relay to hang after the LaunchDarkly client finished initializing, if you had previously queried any relay endpoint _before_ the client finished initializing (e.g. if there was a delay due to network problems and you checked the status resource during that delay).


arun251 (approved):
👍

PR #50, opened 2018-11-27
prepare 5.5.0 release

eli-darkly:
### Added
- The relay now supports additional database integrations: Consul and DynamoDB. As with the existing Redis integration, these can be used to store feature flags that will be read from the same database by a LaunchDarkly SDK client (as described [here](https://docs.launchdarkly.com/v2.0/docs/using-a-persistent-feature-store)), or simply as a persistence mechanism for the relay itself. See `README.MD` for configuration details.
- It is now possible to specify a Redis URL in `$REDIS_URL`, rather than just `$REDIS_HOST` and `$REDIS_PORT`, when running the Relay in Docker. (Thanks, [lukasmrtvy](https://github.com/launchdarkly/ld-relay/pull/48)!)

### Fixed
- When deployed in a Docker container, the relay no longer runs as the root user; instead, it creates a user called `ldr-user`. Also, the configuration file is now in `/ldr` instead of in `/etc`. (Thanks, [sdif](https://github.com/launchdarkly/ld-relay/pull/45)!)


relay.go
```diff
@@ -60,7 +62,8 @@ type EnvConfig struct {
 	ApiKey             string // deprecated, equivalent to SdkKey
 	MobileKey          *string
 	EnvId              *string
-	Prefix             string
+	Prefix             string // used only if Redis, Consul, or DynamoDB is enabled
```
apechimp: Might be better to use pointer types so that you can check for `nil` instead of empty string (`""` seems like it could be a valid prefix).

apechimp: Actually nevermind. This is fine. I misread some of the code below initially 🤦‍♂️

mike-zorn (approved):
Readme and go changes
✨✨✨✨
✨LGTM✨
✨✨✨✨

bash changes seem reasonable, but my review certainly isn't authoritative there.

PR #52, opened 2018-12-18
Fix bash syntax for equality check

t-pascal:
Use correct comparison syntax for entrypoint script.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -17,7 +17,7 @@ port = 8030
 heartbeatIntervalSecs = ${HEARTBEAT_INTERVAL:-15}
 " > $CONF_FILE
 
-if [ "$USE_REDIS" = 1 ] || [ "$USE_REDIS" == "true" ]; then
+if [ "$USE_REDIS" == 1 ] || [ "$USE_REDIS" == "true" ]; then
   if [ -z "${REDIS_HOST}" ] && [ -z "${REDIS_PORT}" ] && [ -z "${REDIS_URL}" ]; then
     echo "Choose REDIS_HOST and REDIS_PORT or REDIS_URL"
     exit 1
@@ -56,7 +56,7 @@ url = "${REDIS_URL}"
   fi
 fi
 
-if [ "$USE_DYNAMODB" = 1 ] || [ "$USE_DYNAMODB" == "true" ]; then
+if [ "$USE_DYNAMODB" == 1 ] || [ "$USE_DYNAMODB" == "true" ]; then
   echo "
 [dynamoDB]
 enabled = true
@@ -65,7 +65,7 @@ localTtl = ${CACHE_TTL:-30000}
 " >> $CONF_FILE
 fi
 
-if [ "$USE_CONSUL" = 1 ] || [ "$USE_CONSUL" == "true" ]; then
+if [ "$USE_CONSUL" == 1 ] || [ "$USE_CONSUL" == "true" ]; then
   echo "
 [consul]
 host = \"${CONSUL_HOST:-localhost}\"
@@ -73,7 +73,7 @@ localTtl = ${CACHE_TTL:-30000}
 " >> $CONF_FILE
 fi
 
-if [ "$USE_EVENTS" = 1 ] || [ "$USE_EVENTS" == "true" ]; then
+if [ "$USE_EVENTS" == 1 ] || [ "$USE_EVENTS" == "true" ]; then
   echo "
 [events]
 eventsUri = \"${EVENTS_HOST:-https://events.launchdarkly.com}\"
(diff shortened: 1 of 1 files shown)


eli-darkly:
Perhaps I'm missing something but it seems to me that it should work the same either way; bash isn't supposed to distinguish between `=` and `==` in tests. Are you seeing it not actually working?

kpankonen:
`==` works with the version of `sh` that comes with busybox-1.27.2-r11 (which alpine:3.7 is using), but using `==` with the built-in test command `[` is not POSIX compatible and may not work with certain versions of `sh`.  

For example, it wouldn't work if `/bin/sh` was linked to `dash`:

```
$ docker run -ti --rm debian:stable sh -c 'if [ "test" == "test" ]; then echo true; fi'
sh: 1: [: test: unexpected operator
$ docker run -ti --rm debian:stable sh -c 'if [ "test" = "test" ]; then echo true; fi'
true
```

Related: https://wiki.ubuntu.com/DashAsBinSh#String_comparison_with_.3D.3D_in_test

t-pascal:
I can go ahead and close this. We had about 3 people stare and scratch our heads at this one, and no one was sure whether it was working as intended, by design, or accident. The original problem is related to the documentation: Sometimes the documentation wants a type NUMBER like 1 or 0 to enable, others take the string (misnamed BOOLEAN) "true" or "false". The code accepts both. Also, USE_REDIS=1 and USE_DYNAMODB=1 only actually enables USE_REDIS... This pull request does not really address any of those issues.

Thanks for the education, and hopefully we can clear up the docs. I have no problems retracting this pull request.

eli-darkly:
Well, based on what @kpankonen mentioned it seems like we should fix this anyway, and the documentation should be clarified. There's another release of ld-relay coming soon so we can address this. Thanks both.

However, @t-pascal, on this—"USE_REDIS=1 and USE_DYNAMODB=1 only actually enables USE_REDIS"—could you clarify what you would expect the behavior to be in that case? There's no way for the relay to actually use two databases at once; are you just saying it should treat that as an error condition? That's not just an issue in the Docker script, it also applies to the configuration file—and to other properties that don't make sense together, e.g. if you set `REDIS_HOST`, `REDIS_PORT`, _and_ `REDIS_URL`, then only the first two are used.

kpankonen:
Sorry, my earlier comment is a little confusing.  Using `=` instead of `==` is more portable, so no change is required.

eli-darkly:
@kpankonen Well, the thing is that right now we have a mix, i.e. there are also some `==` comparisons inside single brackets. And the whole reason we're using single brackets is that this is a sh script and not a bash script (I forgot that when I made my first comment) so I think your point does apply.

t-pascal:
Regarding my comment on using two backends: the feature request would be to support two different clients with one relay: say, Ruby clients talking to a Redis cache and Lambda clients talking to DynamoDB. The proxy itself would use the "fastest" option for itself: memory first, Redis second, DyanmoDB third... It definitely gets a bit hairy quickly. I'm working separately with our support contacts to go over it. A little bit of cleanup in docs and maybe some helpful hints in the entrypoint script to quickly identify the problem helps. Reporting on errors in the config file would be a small but important win.

Regarding single or double-equals: I'll leave it up to you to fix. It sounds like you have it well in hand. I have committed to our support contacts to send you documentation updates so I'll send over a PR that would be mostly advisory. You could accept the PR or rewrite it the way you or your lawyers like it.

eli-darkly:
> the feature request would be to support two different clients with one relay: say, Ruby clients talking to a Redis cache and Lambda clients talking to DynamoDB. The proxy itself would use the "fastest" option for itself: memory first, Redis second, DyanmoDB third

I'm still a bit confused about the distinction you're making between what the clients are "talking to",  and what the proxy is using "for itself."

There are basically two ways the client-proxy relationship can work. One is that the proxy connects to LaunchDarkly to get flags, and uses a feature store (which could be a database) to hold the data that it gets from LD... and then the clients connect to the proxy to get flags. In this scenario, the clients have their own feature stores which are completely separate; they can be in-memory, or use any kind of supported database—it's nothing to do with the proxy's configuration. The proxy doesn't know or care whether the clients are using a database to persist _their own_ state.

The second way is LDD or "daemon" mode, where the clients do not connect to the proxy, nor to LD; they just get flags directly out of the database. So the clients and the proxy have the same feature store. I'm not sure if this is what you're referring to but if so, there's no such thing as "what the proxy is using for itself" in this scenario—it _must_ use what the clients are using in order for any of this to work.

eli-darkly:
FYI, version 5.5.1 was just released and includes the `sh` syntax fix, some clearer (I hope) documentation phrasing, and a validity check on the configuration so that you can't try to enable more than one database.

t-pascal:
That's great news, I was just able to put the finishing touches on using ld-relay in our environment and hope to get it going in production today if possible. I apologise for the confusion about how the proxy and clients work, I'm still learning, and in some cases, having to check the code to see what is happening. Thanks for helping out!

eli-darkly:
@t-pascal Great - please let us know if any other issues come up.

PR #59, opened 2019-04-27
Updating the README

fluffy-dutchman:
Updated the README.md file to include an area about integrating the relay into your own application without using the LoadConfigFile helper method.


eli-darkly:
Sorry, I should have clarified that we were already working on that same update and it will go out in the next release.

PR #65, opened 2019-07-18
split out config structs

mightyguava:
So that `Relay` can be configured programmatically, when used as a library.


bwoskow-ld:
Seems reasonable to me. Thanks for the contribution!

bwoskow-ld (approved):

eli-darkly:
This change is included in the 5.6.0 release.

Issue #62, opened 2019-05-28
SSL support


crivabene:
Hello,

Would it be possible to add the possibility to expose an HTTPS endpoint?

Thanks,
claudio

bwoskow-ld:
Hi Claudio,

Thanks for reaching out with the suggestion. LD Relay currently does not have an HTTPS listener. In the meantime, if you want to interact with LD Relay over HTTPS we recommend that you either use a reverse proxy or put a load balancer in front of the relay.

eli-darkly:
We've finally implemented this in version 5.6.0.

PR #67, opened 2019-07-21
Add TLS Support

InTheCloudDan:
This adds TLS support to the relay exposed endpoint, closing #62.

It also moves the listener to a go func so the listening port properly logs.

### DIFF
--- a/cmd/ld-relay/ld-relay.go
+++ b/cmd/ld-relay/ld-relay.go
@@ -38,7 +38,22 @@ func main() {
 		logging.Error.Printf("Error initializing metrics: %s", err)
 	}
 
-	err = http.ListenAndServe(fmt.Sprintf(":%d", c.Main.Port), r)
+	if c.Main.TLS {
+		if c.Main.TLSCert == "" {
+			logging.Error.Printf("TLS: tlscert required")
+			os.Exit(1)
+		}
+		if c.Main.TLSKey == "" {
+			logging.Error.Printf("TLS: tlskey required")
+			os.Exit(1)
+		}
+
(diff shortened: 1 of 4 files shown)


InTheCloudDan:
Should this also have environment variables `TLS_CERT` and `TLS_KEY` added to the docker-entrypoint.sh?

InTheCloudDan:
I ended up adding the docker-entrypoint configuration. There does end up being a blank line in the main conf section due to using `echo` which is POSIX may not support `-n` vs `printf` but I didn't want to change that if it wasn't needed.

cmd/ld-relay/ld-relay.go
```diff
@@ -38,15 +39,35 @@ func main() {
...
+	}
+
 	if err != nil {
 		if c.Main.ExitOnError {
 			logging.Error.Fatalf("Error starting http listener on port: %d  %s", c.Main.Port, err)
 		}
 		logging.Error.Printf("Error starting http listener on port: %d  %s", c.Main.Port, err)
 	} else {
-		logging.Info.Printf("Listening on port %d\n", c.Main.Port)
+		logging.Info.Printf("TLS Enabled: %s, Listening on port %d\n", strconv.FormatBool(c.Main.TLSEnabled), c.Main.Port)
```
eli-darkly, line 67: I'm not crazy about this wording: I know what you mean by "`TLS Enabled: false`" but at first glance it makes it look like TLS _is_ enabled. Since this option will probably be rarely used, I'd rather just not mention TLS at all if it's not enabled.

InTheCloudDan: I can make it an additional logged line if it is enabled, does that work? I was trying to figure out how to add it as unobtrusively as possible but for it to be logged.

eli-darkly: I think an extra line is fine.

PR #74, opened 2019-09-16
Automatically updating dependencies in docker build

e96wic:


bwoskow-ld:
Hi @e96wic,

Thanks for your submission!

In general, I have a concern about modifying the Dockerfile such that it can automatically update dependencies. With this change, it would theoretically be possible for the Docker image to work properly for some period of time, and then at a later time (after some dependency updates were detected and installed), for the image to stop functioning correctly. I would much prefer for dependencies to be updated through upgrading to a newer base image.

Is there a specific dependency you're looking to upgrade, and if so, which one and why?

Cheers,
Ben

e96wic:
Hi @bwoskow-ld,

our internal monitoring showed that the 5.6.1 image has 3 vulnerabilities:
musl (used in musl-utils, musl) version 1.1.18-r3 has 1 vulnerability (CVE-2019-14697)
busybox (used in ssl_client, busybox) version 1.27.2-r11 has 2 vulnerabilities (CVE-2018-20679)

I managed to fix the first one with this commit. I just saw that you're using alpine as base image for the final container. I overlooked that when I created the PR. I'll give updating that one a shot!

e96wic:
That actually fixed all the security issues. :)

bwoskow-ld:
Excellent! The updated changes are more in line with what I'd expect.

We currently have a test failure on our Go 1.8 CI branch due to a known issue -- we're using a dependency with an unpinned version, and the latest version of that dependency introduced an incompatibility with Go 1.8. As soon as we fix that issue, we'll be able to verify that bumping to the new alpine base image version is fine and move forward with your pull request.

ccgagnon:
Hi, is this alpine image version bump will go in the next release soon? CVE-2019-14697 is big concern for us and we will not use this in prod until it's fixed.

bwoskow-ld:
@ccgagnon -- this was released in [5.7.0](https://github.com/launchdarkly/ld-relay/releases/tag/5.7.0).

ccgagnon:
@bwoskow-ld Our internal monitoring still detect CVE-2019-14697 in the latest 5.8.0 image. Does Dockerfile.goreleaser should also be updated with alpine 3.10.2?

bwoskow-ld:
@ccgagnon you're right -- this should be updated in `Dockerfile.goreleaser` for the change to take effect. This is the dockerfile used for released artifacts. The dockerfile updated in this pull request is only used for development purposes.

I'll commit this change and will include it in the next Relay release.

bwoskow-ld:
@ccgagnon this is fixed in [5.8.1](https://github.com/launchdarkly/ld-relay/releases/tag/5.8.1).

ccgagnon:
Thx @bwoskow-ld!

PR #84, opened 2020-01-03
prepare 5.9.1 release

LaunchDarklyCI:
### Fixed:
- When forwarding events from the PHP SDK, Relay was not preserving additional information related to experimentation features (supported in PHP SDK 3.6.0 and above). As a result, some flag rules might be included in experimentation data in the LaunchDarkly UI when those rules were not selected to be included. Events from other SDKs were not affected.


atrakh (approved):
LGTM when the CI passes

PR #100, opened 2020-08-27
Changes to README, instructions for adding as a dependency and using …

nikunjy:
This change will not impact any kind of compatibility issues. Nothing to declare here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -529,11 +529,24 @@ $ sc create ld-relay DisplayName="LaunchDarkly Relay Proxy" start="auto" binPath
 
 
 ## Integrating the Relay Proxy into your own application
+You will have to import this package as a dependency. 
+```
+go get gopkg.in/launchdarkly/ld-relay.v5
+```
+
+Please note that this package at the time of this writing doesn't use v5 golang client. 
+You might have to also do this 
+```
+go get gopkg.in/launchdarkly/go-server-sdk.v4
+```
 
 You can also use the Relay Proxy to handle endpoints in your own application if you don't want to use the default `ld-relay` application.  Below is an
 example using [Gorilla](https://github.com/gorilla/mux) of how you might instantiate a relay inside your web server beneath a path called "/relay":
 
 ```go
+import (
+    relay "gopkg.in/launchdarkly/ld-relay.v5"
+)
 router := mux.NewRouter()
 configFileName := "path/to/my-config-file"
 cfg := relay.DefaultConfig
--- a/README.md
+++ b/README.md
@@ -533,13 +533,6 @@ You will have to import this package as a dependency.
 ```
 go get gopkg.in/launchdarkly/ld-relay.v5
 ```
-
-Please note that this package at the time of this writing doesn't use v5 golang client. 
-You might have to also do this 
-```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -529,11 +529,24 @@ $ sc create ld-relay DisplayName="LaunchDarkly Relay Proxy" start="auto" binPath
 
 
 ## Integrating the Relay Proxy into your own application
+You will have to import this package as a dependency. 
+```
+go get gopkg.in/launchdarkly/ld-relay.v5
+```
+
+Please note that this package at the time of this writing doesn't use v5 golang client. 
+You might have to also do this
```
eli-darkly: I'm still unclear on why a separate `go get` would be required for something that is a transitive dependency of the package you already imported. Also not sure of the relevance of saying that it doesn't use the v5 SDK; an app that imports Relay would not normally have any interaction with Relay's use of the SDK.

nikunjy: Do you think adding this information will be confusing ?

eli-darkly: Well, _I_ am confused... so I'd like to have a better understanding of what the importance of this information is in your opinion, in case I'm missing something. I realize I didn't put question marks on those two sentences, but I meant them as questions.

eli-darkly: The part about the correct `gopkg.in` import path for the package is of course important— I wasn't questioning that.

eli-darkly: Maybe I should ask more specifically: did you find that if you did _not_ do a separate `go get` for the SDK package, something did not work? If so, what?

sumuis:
Я добавил всех участников.

README.md
nikunjy: In my case it didn't work out of the box without doing the go get. Maybe it is because I was using the launchdarkly client already.

eli-darkly: Well, when I asked "if so, *what* [did not work]?", I was hoping to get a little more information than "it didn't work". Like, what kind of errors did you see? Again, I'm reluctant to add advice to this readme that I wouldn't be able to explain the reason for to anyone— and if this really is a requirement under some circumstances, knowing more specifically what you saw could be helpful to us in providing support to other developers.

nikunjy: Removed the client info. I don't remember the error I had, it was something to do with gomod getting confused between v4 and v5 because I was already using the client.

eli-darkly: > it was something to do with gomod getting confused between v4 and v5 because I was already using the client

I suspect that this may have meant you were using an incorrect import path somewhere. As far as Go is concerned, `gopkg.in/launchdarkly/go-server-sdk.v4` and `gopkg.in/launchdarkly/go-server-sdk.v5` are two completely different and unrelated packages.

eli-darkly:
I'm going to reject this PR because at this point it consists only of adding an import path to the sample code in the readme— which I will do on the v5 branch in case we do any more maintenance releases of v5, and it has already been done in the revised docs for the upcoming v6 release— and adding a statement that tells people to use `go get`, which is not necessarily correct advice depending on what package manager they are using, but is also just repeating very basic Go programming concepts (if you import a package, you will need some way to make the package available, there is nothing special about the Relay code in that regard).

Issue #127, opened 2021-02-03
Relay Proxy DynamoDB Access with AWS IRSA


juanjbrown:
**Is your feature request related to a problem? Please describe.**
Hi. We'd like to run the Relay Proxy in EKS with DynamoDB through AWS's [IRSA (IAM roles for service accounts)](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html), which requires at minimum version 1.23.13 (https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts-minimum-sdk.html).

**Describe the solution you'd like**
We believe that upgrading the AWS SDK to an IRSA support version would address this.

**Describe alternatives you've considered**
* Pull request upgrading the AWS SDK version
* Deploying the Relay Proxy with a sidecar pod that works with IRSA and mounts the role session credentials in a way the Relay Proxy's current SDK version can use

eli-darkly:
Thanks for bringing this to our attention. This should be an easy update, but please see my comment on the PR regarding the choice of version.

eli-darkly:
We've released version 6.1.5 with this update. Please let us know when you've had a chance to try it.

PR #128, opened 2021-02-03
upgrades the Go AWS SDK to v1.23.22 for AWS IRSA support

juanjbrown:
**Requirements**

- [x] I have validated my changes against all supported platform versions

**Related issues**

https://github.com/launchdarkly/ld-relay/issues/127

Upgrades AWS SDK version to the latest bug fix version of v1.23 for AWS IRSA support.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/DataDog/datadog-go v3.7.2+incompatible // indirect
 	github.com/DataDog/opencensus-go-exporter-datadog v0.0.0-20190726131236-c7b72e836616
 	github.com/antihax/optional v1.0.0
-	github.com/aws/aws-sdk-go v1.15.74
+	github.com/aws/aws-sdk-go v1.23.22
 	github.com/fsnotify/fsnotify v1.4.7
 	github.com/gomodule/redigo v1.8.2
 	github.com/googleapis/gax-go v2.0.0+incompatible // indirect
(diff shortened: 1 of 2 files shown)


eli-darkly:
Thanks for the PR. I think we may go a slightly different route by updating the AWS SDK version in the underlying integration library, https://github.com/launchdarkly/go-server-sdk-dynamodb, and then updating the version of that library in `ld-relay` - same result, except it will also benefit anyone who's using our Go SDK.

I'm inclined to update it to the latest AWS SDK release, 1.37.2, rather than 1.23.22, as it seems that they are following semantic versioning and not making breaking changes across minor versions. Are you aware of any reason not to do that? The code seems to work fine with 1.37.2 in our integration tests, and there are other bugfixes they've made along the way that might benefit Relay Proxy users.

juanjbrown:
@eli-darkly Thanks for the quick reply.

Yeah, AWS seems to be good at following semver.  I was trying to make the minimum changes needed just in case 😅, but I can upgrade the SDK in the go-server-sdk-dynamodb project to 1.37.2 and then update its version and the AWS SDK here

eli-darkly:
@juanjbrown 
> but I can upgrade the SDK in the go-server-sdk-dynamodb project to 1.37.2 and then update its version and the AWS SDK here

You don't need to do anything, we're already on it - https://github.com/launchdarkly/go-server-sdk-dynamodb/pull/3

PR #134, opened 2021-04-07
Update in-app README so that routes work as expected

### LINKED ISSUE #133: Help with running ld-relay as part of an application
**Describe the bug**
Could you please tell me what I'm doing wrong? I'm trying to build ld-relay within an application.

**To reproduce**
```
func main() {
	r, err := relay.NewRelay(createRelayConfig(), ldlog.NewDefaultLoggers(), nil)
	if err != nil {
		log.Fatal(fmt.Sprintf("Error creating relay: %s", err))
	}

	router := mux.NewRouter()
	// router.HandleFunc("/test", handleFunc)
	router.PathPrefix("/relay").Handler(r)

	srv := &http.Server{
		Handler: router,
		Addr:    "127.0.0.1:8030",
	}
	srv.ListenAndServe()
}

func createRelayConfig() config.Config {
	var cfg config.Config
	cfg.Main.Port, _ = configtypes.NewOptIntGreaterThanZero(8030)
	cfg.Environment = map[string]*config.EnvConfig{
		"test": {
			SDKKey: config.SDKKey("-redacted-"),
		},
	}
	return cfg
}
```

**Expected behavior**
Expect `curl http://localhost:8030/relay/status` not to return a 404

simonkotwicz:
closes https://github.com/launchdarkly/ld-relay/issues/133

### DIFF
--- a/docs/in-app.md
+++ b/docs/in-app.md
@@ -28,7 +28,7 @@ if err != nil {
 }
 
 router := mux.NewRouter()
-router.PathPrefix("/relay").Handler(r)
+router.PathPrefix("/relay").Handler(http.StripPrefix("/relay", r))
 ```
 
 The above example uses a configuration file. You can also pass in a `config.Config` struct that you have filled in directly. Note that some of the fields use types from `github.com/launchdarkly/go-configtypes` to enforce validation rules.


eli-darkly:
Thanks. Until you submitted this, I had completely forgotten that adding a path prefix of `/relay` was something that was in our example code, and not just something you had decided to do in your own application. I think we should probably update this to show examples both with and without a prefix, because _not_ using one (and just putting the Relay endpoint on a different port from whatever other endpoints the application might have) may be a somewhat more common use case. But in the meantime this makes the current example less wrong.

eli-darkly (approved):

PR #185, opened 2022-04-29
fix empty response body when allowedOrigins is set

joshuaeilers:
**Requirements**

**Related issues**

When setting the allowedOrigins to something other than `*`, as in a real domain, we do not see any response body for `GET` requests. This change stops the short-circuiting that was happening if an origin was matched.

I thought about always passing `*` in production as an absolute last resort.

**Additional context**

### DIFF
--- a/internal/core/middleware/middleware.go
+++ b/internal/core/middleware/middleware.go
@@ -129,14 +129,14 @@ func CORS(next http.Handler) http.Handler {
 			headers = corsContext.AllowedHeaders()
 		}
 		if len(domains) > 0 {
+			domain := domains[0]
 			for _, d := range domains {
 				if r.Header.Get("Origin") == d {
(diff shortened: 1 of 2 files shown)


joshuaeilers:
I'm not sure why the workflow is failing with 401's. Could someone from the LD team help me out with that?

eli-darkly:
> I'm not sure why the workflow is failing with 401's. Could someone from the LD team help me out with that?

The integration test job requires a credential to connect to LD, and that credential is configured as a secret in CircleCI for our own projects - so I think it cannot work in a fork. In other words, this failure isn't related to your change.

eli-darkly (approved):
Thanks for finding and fixing this. The change looks correct, and we should be able to release a patch soon.

eli-darkly:
We've released v6.7.6 with this fix - thanks again.

PR #189, opened 2022-05-20
Cache the replay event in case we get multiple new client connections

moshegood:
**Requirements**

**Related issues**

**Additional context**

### DIFF
--- a/internal/core/streams/stream_provider_server_side.go
+++ b/internal/core/streams/stream_provider_server_side.go
@@ -27,6 +27,10 @@ type serverSideEnvStreamProvider struct {
 type serverSideEnvStreamRepository struct {
 	store   EnvStoreQueries
 	loggers ldlog.Loggers
+
+	mu            sync.Mutex
+	cacheEvent    eventsource.Event
+	eventChecksum string
 }
 
 func (s *serverSideStreamProvider) Handler(credential config.SDKCredential) http.HandlerFunc {
@@ -96,18 +100,35 @@ func (r *serverSideEnvStreamRepository) Replay(channel, id string) chan eventsou
 
 		if err != nil {
 			r.loggers.Errorf("Error getting all flags: %s\n", err.Error())
-		} else {
-			segments, err := r.store.GetAll(ldstoreimpl.Segments())
-			if err != nil {
-				r.loggers.Errorf("Error getting all segments: %s\n", err.Error())
-			} else {
-				allData := []ldstoretypes.Collection{
-					{Kind: ldstoreimpl.Features(), Items: removeDeleted(flags)},
-					{Kind: ldstoreimpl.Segments(), Items: removeDeleted(segments)},
-				}
-				out <- MakeServerSidePutEvent(allData)
-			}
+			return
 		}
+		segments, err := r.store.GetAll(ldstoreimpl.Segments())
+		if err != nil {
+			r.loggers.Errorf("Error getting all segments: %s\n", err.Error())
+			return
+		}
+
+		flagsChecksum := hashKeyedItemDescriptors(flags)
+		segmentsChecksum := hashKeyedItemDescriptors(segments)
+		eventChecksum := flagsChecksum + segmentsChecksum
+
+		r.mu.Lock()
+		defer r.mu.Unlock()
+		if r.eventChecksum == eventChecksum {
+			out <- r.cacheEvent
+			return
+		}
+
+		allData := []ldstoretypes.Collection{
+			{Kind: ldstoreimpl.Features(), Items: removeDeleted(flags)},
+			{Kind: ldstoreimpl.Segments(), Items: removeDeleted(segments)},
+		}
+		event := MakeServerSidePutEvent(allData)
+
+		r.eventChecksum = eventChecksum
+		r.cacheEvent = event
+
+		out <- event
 	}()
 	return out
 }
--- a/internal/core/streams/stream_provider_server_side.go
+++ b/internal/core/streams/stream_provider_server_side.go
@@ -5,6 +5,7 @@ import (
 	"sync"
 
 	"github.com/launchdarkly/ld-relay/v6/config"
+	"golang.org/x/sync/singleflight"
 
 	"github.com/launchdarkly/eventsource"
 	"gopkg.in/launchdarkly/go-sdk-common.v2/ldlog"
@@ -28,9 +29,7 @@ type serverSideEnvStreamRepository struct {
 	store   EnvStoreQueries
 	loggers ldlog.Loggers
 
-	mu            sync.Mutex
-	cacheEvent    eventsource.Event
-	eventChecksum string
+	flightGroup singleflight.Group
 }
 
 func (s *serverSideStreamProvider) Handler(credential config.SDKCredential) http.HandlerFunc {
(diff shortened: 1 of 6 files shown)


moshegood:
Problem: 
When a lot of client connect all at once to a relay, we create a Replay event for each client. These events can use a lot of memory if the underlying project is large. If we have N clients connecting all at once, and the project i M bytes, then we use a total of (M*N) bytes. 

Solution:
We can cache the Replay event as long as the underlying data store returns identical data between calls. This would use O(M) memory. If we have many clients connecting all at once, they can all be served from the same cached event, thereby using O(M) memory total rather than O(N*M). The current solution makes no assumptions about what might change the underlying data storage.

eli-darkly:
I agree that it'd be desirable to be more efficient in a scenario like this. We probably will not be able to review this definitively for a couple of weeks— it's not a huge amount of code but it is a significant implementation change, so as the primary maintainer of this code I would want to look at it in detail, and I'm about to be out for two weeks. I'll comment on a couple of things that jumped out at me so far.

internal/core/streams/stream_provider_server_side_flags.go
```diff
@@ -91,16 +97,44 @@ func (r *serverSideFlagsOnlyEnvStreamRepository) Replay(channel, id string) chan
...
+			[]ldstoretypes.Collection{{Kind: ldstoreimpl.Features(), Items: removeDeleted(flags)}})
+		r.flagsChecksum = checksum
+		r.cacheEvent = event
+
+		out <- event
 	}()
 	return out
 }
+
+func hashKeyedItemDescriptors(flags []ldstoretypes.KeyedItemDescriptor) string {
```
eli-darkly: The parameter name is a bit misleading since it's not necessarily flags.

internal/core/streams/stream_provider_server_side_flags.go
```diff
@@ -91,16 +97,44 @@ func (r *serverSideFlagsOnlyEnvStreamRepository) Replay(channel, id string) chan
...
+	h := sha256.New()
+	for _, flag := range flags {
```
eli-darkly: You'll want to sort these by key first for determinacy, because there's no guarantee that the items always come out of the data store in the same order.

internal/core/streams/stream_provider_server_side.go
```diff
@@ -27,6 +27,11 @@ type serverSideEnvStreamProvider struct {
 type serverSideEnvStreamRepository struct {
 	store   EnvStoreQueries
 	loggers ldlog.Loggers
+
+	mu               sync.Mutex
+	cacheEvent       eventsource.Event
+	flagsChecksum    string
+	segmentsChecksum string
```
eli-darkly: I don't think you actually need to keep track of these two checksums separately— it would be OK to compute just one for the whole data set (as long as the ordering of both the kinds and the keys is determinate).

I would guess that the reason you did it this way is that there's a "flags only" endpoint, so it seemed logical to have the caching of that one depend only on the flags. But in practice that's not a significant issue, because (as you would have no way of knowing; we need to revisit this and either document it better or remove it) the "flags only" endpoint is obsolete; it was used by older SDKs that used an older data model, and those are all now extremely unsupported. No current SDK uses it, and if anyone was connecting to that endpoint for some unusual purpose, the chance that we might not cache it as effectively as possible is not (IMO) worth the extra complication here.

Besides, it's not for sure that flags and segments will always be the only data kinds, so we would like to keep as much of the implementation non-kind-specific as possible— that's the reason the data abstraction in general is designed the way it is.

eli-darkly (changes requested):

internal/core/streams/stream_provider_server_side.go
```diff
@@ -96,18 +100,35 @@ func (r *serverSideEnvStreamRepository) Replay(channel, id string) chan eventsou
...
+		}
+
+		flagsChecksum := hashKeyedItemDescriptors(flags)
+		segmentsChecksum := hashKeyedItemDescriptors(segments)
+		eventChecksum := flagsChecksum + segmentsChecksum
+
+		r.mu.Lock()
+		defer r.mu.Unlock()
+		if r.eventChecksum == eventChecksum {
+			out <- r.cacheEvent
```
eli-darkly: I think you probably want a nil check here. It looks to me like when a `serverSideEnvStreamRepository` is first created, `r.eventChecksum` will be zero and `r.cacheEvent` will be nil. And then if the computed checksum here is also zero, you will end up pushing a nil reference into `out`, which I don't think is valid.

eli-darkly:
I'm sorry that I won't be able to review this further until a couple of weeks from now, but I did have one additional thought in the meantime.

If I understand correctly, this is caching the put event indefinitely— once `r.cacheEvent` is set, it is never cleared, it will only be replaced with a new event if the underlying data changes. So, while this logic does avoid unnecessary _transient_ memory allocation if multiple clients are connecting at once, it increases overall memory usage in the long run: if you have a 20MB environment, then the Relay Proxy will always be holding a 20MB copy of that data (at least, if any clients have ever connected), over and above the memory that is already being used by that data in the regular data store.

One way to avoid that might be to use `singleflight` instead of a cache. That would ensure that the event would only be computed once if multiple clients connect at once, but it would not be retained after the event has been sent.

internal/core/streams/stream_provider_server_side.go
moshegood: The md5 checksum is always 16 bytes long. It will never match `""`.

moshegood:
The Replay function itself doesn't seem like a good candidate for singleflight. The GetAll function, on the other hand, does. I think using the cache and singleflight together would be even better.

moshegood:
I've dropped the cache for now to ensure we go back to the same steady state memory usage.
When multiple clients connect all at once, we will use singleflight to give them all the same event.

internal/core/streams/stream_testdata_test.go
```diff
@@ -49,13 +50,17 @@ type mockStoreQueries struct {
 	fakeSegmentsError error
 	flags             []ldstoretypes.KeyedItemDescriptor
 	segments          []ldstoretypes.KeyedItemDescriptor
+	latency           time.Duration
```
moshegood, line 53: Added this for testing concurrent calls to Replay - to ensure we have both calls use the same flightgroup.

eli-darkly (approved):
I think this is a desirable change and the current state of the PR looks good to me. However, I'd like to take a second pass at it in terms of testing before release (I think the existing tests are correct, but their style is a bit different than how the rest of our test code is written and we may also want more)— so I'm going to merge it into a `contrib` branch first (since in this repository we normally try to have the default branch match what's in the latest release). We will treat this as a high-priority fix due to its potential benefit in terms of reducing transient memory usage spikes. Thanks for putting so much thought into this.

internal/core/streams/stream_provider_server_side_test.go
```diff
@@ -202,4 +203,85 @@ func TestStreamProviderServerSide(t *testing.T) {
...
+				}
+				if expectedData != actual.Data() {
+					t.Errorf("Replay() = %v; wanted %v", actual.Data(), expectedData)
+				}
+			case <-time.After(time.Second):
+				t.Error("expected an event to be returned")
+			}
+
+		})
+		t.Run("ConcurrentCalls", func(t *testing.T) {
```
eli-darkly, line 246: As part of the unit test polishing work I'm doing, I'm changing how this one works because I think there's a logic error in it.

The test on line 282 does not, I think, actually verify that the computation was not done twice: `event1 != event2` is comparing two interface values, and in Go that is not a _reference_ equality test (as in, "are they literally the same instance of the type"), it is a test of whether they have the same concrete types and whether their values as those concrete types are equal. In other words, if `a` and `b` are separately created instances of some struct type and their fields have the same values, and you cast them both to some interface type, the resulting interface values are equal. And since the data store in this case returns the same flag/segment data every time it is queried (the same _empty_ data, since it was created with flags and segments of `nil`), this test can't possibly fail. So I'm implementing this differently, using a store that is instrumented to return different data for each query.

eli-darkly: Sorry, my previous comment was not quite right. The test as written does work, in that if I remove the actual fix and run the new test, the test does fail— it detects inequality of the events on line 282. However, that is because of an implementation detail that I'd rather not rely on: the use of the `streams.deferredEvent` type, which memoizes the value of the event. That forces Go to treat the equality test as reference equality because the type contains things that can't be compared by value. But if the concrete types of `event1` and `event2` were a simpler struct type that just had fields for the event name and data, then line 282 would not work. I'd rather have this test be independent of such implementation details that aren't directly relevant to the logic we're testing, so I still think it's best to use an instrumented store as I described.

eli-darkly:
This improvement is included in the [v6.7.9](https://github.com/launchdarkly/ld-relay/releases/tag/v6.7.9) release.

Issue #280, opened 2024-01-03
Atomic update to an offline configuration archive causes subsequent changes to be missed


gmckerrell:
**Describe the bug**
When operating in "offline" mode, the relay proxy watches for configuration archive changes.
The nsnotify project explicitly recommends that directories should be watched not specific files
https://pkg.go.dev/github.com/fsnotify/fsnotify#NewWatcher
> Watching individual files (rather than directories) is generally not recommended as many programs (especially editors) update files atomically: it will write to a temporary file which is then moved to to destination, overwriting the original (or some variant thereof). The watcher on the original file is now lost, as that no longer exists.

However currently the relay proxy watches the explicit archive file path.

This means that an atomic update will work exactly once; the `CHMOD` and  `REMOVED` events get signaled and the appropriate update occurs, however the removal of the file being watched means that subsequent changes are not seen.

**To reproduce**
Start up relay proxy using `offline/config.tar.gz` as the offline configuation.
```
$ cp config1.tar.gz offline/config.tar.gz.tmp
$ mv offline/config.tar.gz.tmp offline/
```
config1 details are reloaded
```
$ cp config2.tar.gz offline/config.tar.gz.tmp
$ mv offline/config.tar.gz.tmp offline/
```
config2 details are *not* reloaded

**Expected behavior**
atomic moving of configuration archive into place works

**Logs**
The following test logs indicates the failure mode
![image](https://github.com/launchdarkly/ld-relay/assets/31652935/bcd7c087-a3d1-4305-8481-cd13a3aef7c1)
```golang
func TestFileAtomicUpdatedWithValidDataAddedEnvironment(t *testing.T) {
	archiveManagerTest(t, func(filePath string) {
		writeAtomicArchive(t, filePath, false, nil, testEnv1)
	}, func(p archiveManagerTestParams) {
		require.NoError(t, p.archiveManagerError)

		p.expectEnvironmentsAdded(testEnv1)

		writeAtomicArchive(t, p.filePath, false, nil, testEnv1, testEnv2)

		p.expectEnvironmentsAdded(testEnv2)
		p.expectReloaded()

		writeAtomicArchive(t, p.filePath, false, nil, testEnv1)

		p.expectEnvironmentsDeleted(testEnv2.id())
		p.expectReloaded()
	})
}
```
**SDK version**
8.2.1

**Language version, developer tools**
using the following docker image to build and test
```
FROM golang:alpine

RUN apk add make bash gcc libc-dev
[...]

PR #281, opened 2024-01-03
Modify the archive manager so that atomic updates of the configuration archive work correctly.

gmckerrell:
**Requirements**

**Related issues**
https://github.com/launchdarkly/ld-relay/issues/280

Rather than use `nsnotify` to watch the explicit configuration file, we monitor the parent directory which means both slow-copy and atomic update semantics both work as expected.

In our current use case we have deferred to slow-copy semantics rather than atomic updates.

### DIFF
--- a/internal/filedata/archive_manager.go
+++ b/internal/filedata/archive_manager.go
@@ -2,6 +2,7 @@ package filedata
 
 import (
 	"io"
+	"path/filepath"
 	"os"
 	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


louis-launchdarkly:
Hello @gmckerrell, thank you for your contribution and explanation. We will test the change and will update you after that.

Filed internally as _228550_.

cwaldren-ld:
Hi @gmckerrell, thanks again. I will merge this now and the next Relay v7/v8 releases should have the changes.

PR #300, opened 2024-02-09
feat: build Docker images for ARMv7 and ARM64v8 architectures

cwaldren-ld:
This commit modifies our goreleaser configuration to achieve two goals: 
1) Produce release binaries for ARM/ARM64 architectures 
2) Produce individual docker images for ARM/ARM64, and produce a final docker manifest referencing those images.

BEGIN_COMMIT_OVERRIDE
feat: build Docker images for ARMv7 and ARM64v8 architectures
END_COMMIT_OVERRIDE


.goreleaser.yml
```diff
@@ -59,15 +63,83 @@ release:
...
 
 dockers:
+  # i386
   - image_templates:
-    - "launchdarkly/ld-relay:{{ .Version }}"
-    - "launchdarkly/ld-relay:v{{ .Major }}"
-    - "launchdarkly/ld-relay:latest"
+    - "launchdarkly/ld-relay:{{ .Version }}-i386"
+    - "launchdarkly/ld-relay:v{{ .Major }}-i386"
+    - "launchdarkly/ld-relay:latest-i386"
```
keelerm84, line 70: If we make a new branch and forget to remove this bit, that's going to cause problems when we release older versions right?

keelerm84 (approved):

.goreleaser.yml
cwaldren-ld: (for posterity: discussed offline that it's not optimal that the `latest` tag flip flops. Have another PR up to fix this.)

PR #301, opened 2024-02-09
chore(v7): only tag latest Docker images on default branch

cwaldren-ld:
When a v7 release happens, one of the docker images published will be `ld-relay:latest` (this is also true for the individual architecture images, after the ARM builds PR was just merged.) The same occurs for v8.

This is confusing, to say the least.  You might get a `v7` or `v8` revision just depending on the order of Relay releases. 

Although one shouldn't ever rely on `latest` tag in production, some may, and it's the default when you run `docker pull`.

This PR modifies the publishing step to only push `latest` and `latest-{architecture}` images from the default branch (which is currently `v8`).

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -83,7 +83,7 @@ dockers:
     goos: linux
     goarch: '386'
     dockerfile: Dockerfile.goreleaser
-    skip_push: "{{ if .Env.DEFAULT_BRANCH eq .Major }}false{{ else }}true{{ end }}"
+    skip_push: "{{ .Env.NOT_DEFAULT_BRANCH }}"
     build_flag_templates:
     - "--pull"
     - "--platform=linux/386"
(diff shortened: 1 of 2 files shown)


.goreleaser.yml
```diff
@@ -67,7 +67,6 @@ dockers:
   - image_templates:
     - "launchdarkly/ld-relay:{{ .Version }}-i386"
```
cwaldren-ld, line 68: So, this isn't exactly elegant.. but what I've done is just split out the `latest` tags from each instance of `image_templates`, into a new `image_template` block.

This is because we can only control the `skip_push` argument at the `image_template` level.

The idea is - for each architecture, build *all* the tags (we don't really have a choice here, unless we dynamically generate the config file), but only push the `latest` tag for the architecture if we're on the default branch (via `skip_push` config item).

Then in the manifest section starting on line 169, don't push the `latest` manifest unless we're on the default branch.

More yaml, but this should allow for us to have a safe switchover when we make a new major version and mark it as default.

keelerm84 (approved):

PR #308, opened 2024-02-12
ci: automatically create PR to update Go versions

cwaldren-ld:
There's no real reason not to automatically create a PR whenever the Go version needs updating. If the docker/alpine images aren't available, then we should see a workflow failure.

Test: https://github.com/launchdarkly/ld-relay/pull/311


keelerm84 (approved):
🔥

PR #312, opened 2024-02-12
ci: automatically create PR to update Go versions

cwaldren-ld:
Re-introduces the check-go-versions workflow on v7 branch. Now the workflow creates a PR instead of filing an issue.


kinyoklion:
Is this PR intended to be superseded by the matrix one?

cwaldren-ld:
Yea this is useless since it won't run on the schedule. Will prepare some updates to align all the scheduled workflows to use a matrix of v7 and v8.

PR #317, opened 2024-02-13
feat: replace offline-mode filewatcher with polling

cwaldren-ld:
**Requirements**

 I have validated my changes against all supported platform versions:
- [x] Windows (todo)
- [x] Mac
- [x] Linux

**Related issues**

We've encountered robustness issues with the current offline-mode file-watcher implementation in Docker images that use shared volumes. Additionally, we're seeing intermittent CI failures in the file-watcher tests.

Overall, I believe the complexity introduced by file-watcher isn't worth the payoff. It is supposed to unify all the different operating system methods of notifying that files have changed, but it falls short. 

The vast majority of the latency when using offline mode would be downloading the actual archive from LaunchDarkly  and the interval between those downloads - which might be minutes/hours/days.

This PR replaces the file-watcher with a simple polling loop. Every interval, it calls `stat()` and determines if the offline archive needs to be reloaded. 

The default interval is 1 second, and the minimum configurable is 100ms. The minimum was chosen to protect the system in case of accidental configuration of an extremely short interval (like 0). 

In practice, most users may raise the interval - for example, if they're fetching the archive every hour, there is no need to use a 1 second interval.

An alternative is to more rigorously test the existing file-watcher, document all of its failure modes, and overall invest more in that solution.


internal/filedata/archive_manager_test.go
```diff
@@ -186,7 +186,7 @@ func TestFileUpdatedWithInvalidDataAndDoesNotBecomeValid(t *testing.T) {
 
 		writeMalformedArchive(p.filePath)
```
cwaldren-ld, line 187: TODO: It's unclear to me whether these should be errors or warnings. In Eli's spike of this functionality, there is extra logic to determine if the file once existed and then was removed - which might not be considered an error. 

This could be nice but it's more logic in the loop.

internal/filedata/archive_manager.go
```diff
@@ -100,106 +88,42 @@ func (am *ArchiveManager) Close() error {
...
+				continue
 			}
-
-		case <-retryCh:
-			// If needRetry is false, this is an obsolete signal - we've already successfully reloaded
-			if pendingRetry {
-				am.loggers.Debug("Got retry signal")
-				pendingRetry = false
-				maybeReload()
+			if fileMayHaveChanged(prevInfo, nextInfo) {
```
kinyoklion, line 113: So, one thing to note, this `fileMayhaveChanged` would be a bad match for use with a watcher. If you programmatically modify a file, then it gets several changes in a small period. This period is less than the timestamp precision, which is 1 second typically. So depending on how things go you get random states within that update cycle. Node server had this problem.

I do think there is a letter problem here. Which is 100ms is also faster, multiple updates within 1 second you could load the first change, and then subsequently thetimestamp remains the same, so you don't. But it isn't too bad because it also has to be the same size. So it would affect a change of variation from something like "a" to "b".

internal/filedata/archive_manager_test.go
kinyoklion: If you are using relay proxy is deleting a file a valid thing to do, and what is the expected outcome of it?

kinyoklion (approved):

internal/filedata/archive_manager_test.go
cwaldren-ld: It's valid in the sense that you can do it and nothing bad will happen in that exact moment, since the environments are already loaded into memory. 

But I'd say it's not an expected thing, and not recommended, since it means that if Relay crashes or restarts, it's going to fail.

PR #319, opened 2024-02-13
ci: add v7 branch to matrix for installation tests + security scans

cwaldren-ld:
Sister PR: https://github.com/launchdarkly/ld-relay/pull/318

This adds `v7` into the installation test + security scan scheduled workflows.

### DIFF
--- a/.github/workflows/daily-installation-test.yml
+++ b/.github/workflows/daily-installation-test.yml
@@ -1,20 +1,5 @@
 name: 'Install Relay Go Module'
 on:
-  workflow_dispatch:
-    inputs:
-      relay-version:
-        description: "Relay version to install, e.g. 'latest'"
-        type: string
-        required: false
-      relay-major:
-        description: "Relay major version to install, e.g. 'v8'."
-        type: string
-        required: false
-      branch:
-        description: "Branch to test."
-        type: string
-        required: false
-
   schedule:
     - cron: "0 8 * * *"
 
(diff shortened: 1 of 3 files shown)


.github/workflows/daily-installation-test.yml
```diff
@@ -29,12 +14,12 @@ jobs:
...
     steps:
       - uses: actions/checkout@v4
-        with:
-          ref: ${{ inputs.branch }}
       - uses: ./.github/actions/install-relay
         with:
           go-version: ${{ matrix.go-version }}
-          relay-version: ${{ inputs.relay-version }}
-          relay-major: ${{ inputs.relay-major }}
+          relay-version: 'latest'
```
kinyoklion, line 24: I am confused by the versions.

kinyoklion (approved):

.github/workflows/daily-installation-test.yml
cwaldren-ld: It's confusing. The `v7` or `v8` part is part of the module path, and then you indicate the actual tag. 

So for ex relay's latest release is:
-  `/ld-relay/v8@latest`, or
- `/ld-relay/v8@v8.3.1`

`v7`'s latest release is:
- `/ld-relay/v7@latest` or,
- `/ld-relay/v7@v7.5.0`

So it's really indicating "give me the v7 module (which is totally independent of the v8 module), and then give me a specific tag of that module."

I could perhaps have some better words to use in the yml, but I wasn't sure.

PR #324, opened 2024-02-28
chore: add Backstage catalog file 

dg-ld:
Adds the service catalog file that will be ingested by our Backstage instance.


dje:
Let's move this to the shared manifest (from [slack thread](https://launchdarkly.slack.com/archives/C0682LRE90Q/p1709152751494939)).

Issue #353, opened 2024-04-26
Feature Request - allow configuring which host ld-proxy-relay binds to


brycefisher:
**Is your feature request related to a problem? Please describe.**
When I run ld-relay-proxy on my mac as part of a test suite, I see a security popup because ld-relay-proxy binds to all network interfaces and is nonconfigurable.

**Describe the solution you'd like**
An environment variable that would allow me to force a specific bind address (in my case, only ever 127.0.0.1)

**Describe alternatives you've considered**
I could use the ldtest packages _inside_ my application, but I'm trying to test the fully compiled binary, and I'd rather not change any application source code.

**Additional context**
I think I could probably contribute a small PR to this effect, depending on if any tests needed to be added, etc.

cwaldren-ld:
Hi @brycefisher , this sounds like a reasonable request. If you wouldn't mind contributing a PR, that would be helpful.

PR #354, opened 2024-05-02
feat: add HOST config to control which network interface ld-relay binds to

brycefisher:
**Requirements**

- [ ] I have added test coverage for new or changed functionality
- [ ] I have followed the repository's [pull request submission guidelines](../blob/master/CONTRIBUTING.md#submitting-pull-requests)

**Related issues**

https://github.com/launchdarkly/ld-relay/issues/353

By default, the existing behavior is preserved of binding all to network interfaces. However, if the config HOST is provided, ld-relay will attempt to bind that host on given PORT.

We could offer a boolean flag "BIND_LOCALHOST" or similar that would explicitly set the bind address to `localhost:PORT`. The provided option seemed more inline what I've seen in other projects, but I've no strong personal preference for either.

ALternatively we could allow users to specify a fully formed bind address including port and hostname in a single string. I don't like that option because we'd potentially break how folks configure how they instruct ld-relay where to run if they tried to provide both a fully formed address and a port. Keeping HOST and PORT distinct seems simpler to implement and reason about AFAICT.

**Additional context**

I'm not sure what this new config should be called 🤔

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -133,6 +133,7 @@ type MainConfig struct {
 	StreamURI                  ct.OptURLAbsolute        `conf:"STREAM_URI"`
 	BaseURI                    ct.OptURLAbsolute        `conf:"BASE_URI"`
 	ClientSideBaseURI          ct.OptURLAbsolute        `conf:"CLIENT_SIDE_BASE_URI"`
+	Host                       string                   `conf:"HOST"`
 	Port                       ct.OptIntGreaterThanZero `conf:"PORT"`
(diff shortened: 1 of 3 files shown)


brycefisher:
@cwaldren-ld  is there anything special I need to do to run the existing tests? Is `go test` good enough to iterate locally? Apologies for missing the contribution guide! Feel free to point me at that.

cwaldren-ld:
Hi @brycefisher , the guide is here: https://github.com/launchdarkly/ld-relay/blob/v8/CONTRIBUTING.md

PR #364, opened 2024-05-20
feat: Add support for receiving gzip event payloads

keelerm84:
BEGIN_COMMIT_OVERRIDE
feat: Support receiving compressed event payloads (#364)
feat: Add new `MaxInboundPayloadSize` configuration to limit event payload sizes (#364)
feat: gzip event payloads before sending to upstream LaunchDarkly APIs (#364)
fix: Bump minimum go version to 1.21 (#364)
END_COMMIT_OVERRIDE


config/config.go
```diff
@@ -3,6 +3,7 @@ package config
 import (
 	"time"
 
+	"github.com/alecthomas/units"
```
keelerm84: Adding this library because it allows for a nice size type (e.g. 15GiB).

config/config.go
```diff
@@ -179,11 +180,12 @@ type OfflineModeConfig struct {
...
-	SendEvents    bool                     `conf:"USE_EVENTS"`
-	FlushInterval ct.OptDuration           `conf:"EVENTS_FLUSH_INTERVAL"`
-	Capacity      ct.OptIntGreaterThanZero `conf:"EVENTS_CAPACITY"`
-	InlineUsers   bool                     `conf:"EVENTS_INLINE_USERS"`
+	EventsURI                ct.OptURLAbsolute        `conf:"EVENTS_HOST"`
+	SendEvents               bool                     `conf:"USE_EVENTS"`
+	FlushInterval            ct.OptDuration           `conf:"EVENTS_FLUSH_INTERVAL"`
+	Capacity                 ct.OptIntGreaterThanZero `conf:"EVENTS_CAPACITY"`
+	InlineUsers              bool                     `conf:"EVENTS_INLINE_USERS"`
+	MaxReceivablePayloadSize units.Base2Bytes         `conf:"EVENTS_MAX_RECEIVABLE_PAYLOAD_SIZE"`
```
keelerm84: I would welcome a better name.

internal/util/reader.go
```diff
@@ -0,0 +1,107 @@
...
+
+	wrappedBaseStream *byteCountingReader
+	stream            *io.LimitedReader
+}
+
+// NewReader creates a new reader
+func NewReader(r io.ReadCloser, isGzipped bool, maxBytes int64) (io.ReadCloser, error) {
+	// If this isn't compressed, and we don't want to limit the size, just
+	// return the original reader
+	if !isGzipped && maxBytes == 0 {
```
keelerm84: To maintain backwards compatibility, if the new max payload size isn't specified,  it defaults to 0 which we will interpret to mean no restrictions on the size of the payload (the current behavior).

Because of this, we have a couple of early return chances here -- line 30 and 48. Because we are returning different types, the return type on line 26 is now an `io.ReadCloser` instead of the `PayloadReader` like one might expect.

This isn't exactly optimal. In a future release, when we don't mind breaking backwards compatibility, we won't have to deal with this 0 nonsense, and the logic can basically fall back through.

Alternatively, what do you think about making the config optional? If it isn't set, then we add no restriction, but once you specify a value, we enforce that limit, including 0.

internal/util/reader_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestUncompressed(t *testing.T) {
+	// make sure the bytes read are correct for non-gzipped streams
+	nonZipBytes := []byte("this is a test")
+
+	reader, _ := NewReader(io.NopCloser(bytes.NewReader(nonZipBytes)), false, 1000)
+	payloadReader, _ := reader.(*PayloadReader)
```
keelerm84, line 20: Since we aren't returning the PayloadReader type, we have to do this typehint stuff in a couple of places here.

internal/util/reader_test.go
```diff
@@ -0,0 +1,77 @@
...
+	assert.Equal(t, int64(len(nonZipBytes)), payloadReader.GetUncompressedBytesRead())
+	assert.Equal(t, nonZipBytes, readBytes)
+}
+
+func TestPayloadBytesTracking(t *testing.T) {
+	// build a string
+	var ret string
+	for i := 0; i < 500; i++ {
+		ret += "00"
+	}
```
kinyoklion: `ret := strings.Repeat("0", 1000)`

internal/util/reader.go
kinyoklion: We have some existing payload size in the megabytes that will cause the service to send a 413. It seems like we could reasonably have a default limit that is some size larger than that.

config/config.go
cwaldren-ld: How about: `MaxInboundPayloadSize` and `EVENTS_MAX_INBOUND_PAYLOAD_SIZE`?


If we added a config for the max that could be sent to LD, we could call it `Outbound`.

internal/util/reader.go
cwaldren-ld: I like the idea that to maintain backwards compat, we allow the max payload size to not be set. 

If not set, then there is no restriction: old behavior.

If it is set, then it is restricted to whatever they choose.  We _could_ use -1 or 0 as a sentinel value to indicate "no restriction", but I think it would probably be better to enforce a positive number. That way you can be guaranteed that if you've configured this thing, you are protected to some extent.

cwaldren-ld: Then in a major version, we start setting it automatically to some [well defined but suitable] max value (whatever ryan mentioned?).

config/config.go
keelerm84: Updated to this new name.

internal/util/reader.go
keelerm84: Update the code to use the new configtypes.OptBase2Bytes type. This lets it default to nothing being enforced, but once it is set, it must be > 0 so we can do some simple enforcing.

internal/util/reader_test.go
```diff
@@ -0,0 +1,73 @@
...
+	payloadReader, _ := reader.(*PayloadReader)
+	readBytes, _ := io.ReadAll(reader)
+
+	assert.Equal(t, int64(len(zipBytes)), payloadReader.GetBytesRead())                // compressed is 32
+	assert.Equal(t, int64(len(nonZipBytes)), payloadReader.GetUncompressedBytesRead()) // uncompressed is 1000
+	assert.Equal(t, nonZipBytes, readBytes)
+}
+
+func TestZipBombing(t *testing.T) {
+	// build a string
```
cwaldren-ld: Probably don't need this comment

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/launchdarkly/ld-relay/v8
 
-go 1.20
+go 1.21
```
cwaldren-ld, line 3: Checking that this change actually requires a bump to the minimum go version?

kinyoklion (approved):

PR #376, opened 2024-06-05
ci: add workflow to check for new alpine versions

cwaldren-ld:
This adds a workflow that checks the current version of Alpine used in the Dockerfile against the latest available.

If they differ, it runs the `./scripts/update-alpine-version.sh` script. This will occur on both the v7 and v8 branches, just like the Go version checker.


.github/workflows/check-alpine-version.yml
```diff
@@ -0,0 +1,68 @@
...
+          echo "version=$version" >> $GITHUB_OUTPUT
+
+      - name: Update from ${{ steps.alpine-current.outputs.version }} to alpine:${{ steps.alpine-latest.outputs.version }}
+        id: update-alpine
+        if: steps.alpine-current.outputs.version != steps.alpine-latest.outputs.version
+        run: ./scripts/update-alpine-version.sh ${{ steps.alpine-latest.outputs.version }}
+
+      - name: Create pull request
+        if: steps.update-alpine.outcome == 'success'
+        uses: peter-evans/create-pull-request@v6
```
kinyoklion: We should probably be using a SHA for this one instead of v6.

kinyoklion:
Is the test issue a flake?

cwaldren-ld:
> Is the test issue a flake?

Probably, although I don't think I've seen this specific one flake. Will find out soon enough.

.github/workflows/check-alpine-version.yml
cwaldren-ld: Because it's external in general or something about the specific action? (The Go version action and Trivy scan also use tags)

kinyoklion: Effectively because it is smaller. `peter-evans` versus github/microsoft, etc.

kinyoklion (approved):

PR #401, opened 2024-06-18
feat: periodic key revocation

cwaldren-ld:
This replaces Relay's key rotation system with a new abstraction: `credential.Rotator`

Previously, key rotation was the concern of the stream message handler, project manager, environment manager, and environments. This was because SDK key expiry was handled by the stream manager at the highest level, even though the actual credential handling needed to be delegated to the lowest level (environments.)

Now, key rotation happens locally within an environment. Key deprecation messages are pushed down the stack until they get to an environment, which hands it off to the `Rotator`. 

A `Rotator` can be queried at any time to determine what keys should be added or removed from an environment. It is purely logical and doesn't deal with goroutines or channels, making it easy to test. 

---------

**Behavioral Change**
Key expiration is no longer instantaneous, it now happens on a configurable interval - for example, expired keys are checked every minute. The tradeoff is higher latency, but easier to test and predict the behavior. 

The interval can be controlled with config option `EXPIRED_CREDENTIAL_CLEANUP_INTERVAL`.


config/config_field_types.go
```diff
@@ -44,6 +42,13 @@ type FilterKey string
 // DefaultFilter represents the lack of a filter, meaning a full LaunchDarkly environment.
 const DefaultFilter = FilterKey("")
 
+func last4Chars(s string) string {
```
cwaldren-ld, line 45: TODO: probably make a better masking function, maybe append some '*'s to it?

internal/autoconfig/stream_manager.go
```diff
@@ -307,12 +307,6 @@ func (s *StreamManager) consumeStream(stream *es.Stream) {
 			if shouldRestart {
 				stream.Restart()
 			}
-
```
cwaldren-ld, line 310: Now stream manager doesn't need to worry about keys, which was entirely outside its normal purview.

internal/autoconfig/stream_manager.go
```diff
@@ -329,17 +323,11 @@ func (s *StreamManager) dispatchEnvAction(id config.EnvironmentID, rep envfactor
 		return
 	case ActionInsert:
 		params := rep.ToParams()
```
cwaldren-ld, line 310: No more need to modify the data model before passing it deeper into the handlers.

internal/autoconfig/stream_manager_expiring_key_test.go
```diff
@@ -52,111 +50,112 @@ func expectNoKeyExpiryMessage(p streamManagerTestParams) {
...
-		msg := p.requireMessage()
-		require.NotNil(t, msg.update)
-		assert.Equal(t, testEnv1.ToParams(), *msg.update)
-		assert.Equal(t, config.SDKKey(""), msg.update.ExpiringSDKKey)
-
-		expectNoKeyExpiryMessage(p)
-	})
-}
+//
+//func TestExpiringKeyInPutMessage(t *testing.T) {
```
cwaldren-ld: These tests need to be rewritten elsewhere

internal/envfactory/env_rep.go
```diff
@@ -62,22 +62,33 @@ type ExpiringKeyRep struct {
...
-		MobileKey:      r.MobKey,
-		ExpiringSDKKey: r.SDKKey.Expiring.Value,
-		TTL:            time.Duration(r.DefaultTTL) * time.Minute,
-		SecureMode:     r.SecureMode,
+		SDKKey:     r.SDKKey.Value,
+		MobileKey:  r.MobKey,
+		TTL:        time.Duration(r.DefaultTTL) * time.Minute,
+		SecureMode: r.SecureMode,
+	}
+
```
cwaldren-ld, line 96: Todo: can this be set directly and default values will flow through?

internal/relayenv/env_context.go
```diff
@@ -45,22 +50,9 @@ type EnvContext interface {
...
-	// RemoveCredential removes a credential from the environment. Any active stream connections using that
-	// credential are immediately dropped.
-	//
-	// If the credential is an SDK key, then the SDK client that we started with that SDK key is disposed of.
-	RemoveCredential(credential.SDKCredential)
-
-	// DeprecateCredential marks an existing credential as not being a preferred one, without removing it or
-	// dropping any connections. It will no longer be included in the return value of GetCredentials(). This is
-	// used in Relay Proxy Enterprise when an SDK key is being changed but the old key has not expired yet.
-	DeprecateCredential(credential.SDKCredential)
```
cwaldren-ld, line 63: Rather than speak about adding/removing/deprecating credentials, that level of detail is no longer exposed on the EnvContext interface.

Instead, components can ask the environment to rotate credentials. I've split them out into the actual mobile/env/SDK key types, because deprecation is actually only relevant for SDK keys. This makes it clear.

internal/relayenv/env_context_impl.go
```diff
@@ -73,38 +78,43 @@ type EnvContextImplParams struct {
...
+	sdkInitTimeout            time.Duration
+	metricsManager            *metrics.Manager
+	metricsEnv                *metrics.EnvironmentManager
+	metricsEventPub           events.EventPublisher
+	dataStoreInfo             sdks.DataStoreEnvironmentInfo
+	globalLoggers             ldlog.Loggers
+	ttl                       time.Duration
+	initErr                   error
+	creationTime              time.Time
+	filterKey                 config.FilterKey
```
cwaldren-ld, line 118: The diff here is adding:
- keyRotator
- stop/doneMonitoringCredentials
- connectionMapper (which is really just the Relay struct under the hood, needed a way to execute some callbacks on that struct)

internal/relayenv/env_context_impl.go
```diff
@@ -447,97 +523,29 @@ func (c *envContextImpl) SetIdentifiers(ei EnvIdentifiers) {
...
 	defer c.mu.RUnlock()
-	// There might be multiple clients if there's an expiring SDK key. Find the SDK key that has a true
-	// value in our map (meaning it's not deprecated) and return that client.
-	for cred, valid := range c.credentials {
-		if sdkKey, ok := cred.(config.SDKKey); ok && valid {
-			return c.clients[sdkKey]
-		}
-	}
-	return nil
+	return c.clients[c.keyRotator.SDKKey()]
```
cwaldren-ld, line 563: Todo: could be safer here and return nil if keyRotator.SDKKey() is not defined.

config/config.go
```diff
@@ -134,24 +138,25 @@ type Config struct {
...
+	MaxClientConnectionTime          ct.OptDuration           `conf:"MAX_CLIENT_CONNECTION_TIME"`
+	DisconnectedStatusTime           ct.OptDuration           `conf:"DISCONNECTED_STATUS_TIME"`
+	TLSEnabled                       bool                     `conf:"TLS_ENABLED"`
+	TLSCert                          string                   `conf:"TLS_CERT"`
+	TLSKey                           string                   `conf:"TLS_KEY"`
+	TLSMinVersion                    OptTLSVersion            `conf:"TLS_MIN_VERSION"`
+	LogLevel                         OptLogLevel              `conf:"LOG_LEVEL"`
+	BigSegmentsStaleAsDegraded       bool                     `conf:"BIG_SEGMENTS_STALE_AS_DEGRADED"`
+	BigSegmentsStaleThreshold        ct.OptDuration           `conf:"BIG_SEGMENTS_STALE_THRESHOLD"`
+	ExpiredCredentialCleanupInterval ct.OptDuration           `conf:"EXPIRED_CREDENTIAL_CLEANUP_INTERVAL"`
```
cwaldren-ld, line 159: This last config is the only changed thing.

internal/credential/rotator.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+}
+
+func (r *Rotator) expireSDKKey(sdkKey config.SDKKey) {
+	r.loggers.Infof("Deprecated SDK key %s has expired and is no longer valid for authentication", sdkKey.Masked())
+	delete(r.deprecatedSdkKeys, sdkKey)
+	r.expirations = append(r.expirations, sdkKey)
+}
+
+func (r *Rotator) Query(now time.Time) (additions []SDKCredential, expirations []SDKCredential) {
```
cwaldren-ld: nit: think of a better name for this since query implies nothing is changed internally (idempotent)

louis-launchdarkly (approved):
We pair reviewed this together and the clean up, and encapsulate the add/remove credential as the update function is good.

PR #408, opened 2024-06-25
feat: offline mode key rotation

cwaldren-ld:
This PR adds support for SDK key rotation (both with and without deprecation periods) to offline mode. 

What this means is that whenever an archive is loaded, if there is a deprecated key, it will be accepted for new SDK connections until it reaches its expiration date.


louis-launchdarkly (approved):
Casey and I reviewed this together on Slack.

PR #451, opened 2024-10-02
ci: avoid Trivy action rate limiting

cwaldren-ld:
Trivy's ghcr.io container is globally rate limited. I guess that mean's Trivy is pretty popular! 

This PR switches to a different database location so that we can avoid the ratelimit, which was causing the daily security scans to fail.


pkaeding (approved):
looks like it works, so far! image

PR #452, opened 2024-10-15
feat: add support for client-side prerequisite events

cwaldren-ld:
This PR updates Relay's client-side evaluation logic to track the top-level prerequisites associated with a flag. 

These prerequisites are then passed back in the response as a `prerequisites` key, allowing client-side SDKs to properly emit prerequisite events.


relay/relay_endpoints.go
```diff
@@ -262,6 +270,15 @@ func evaluateAllShared(w http.ResponseWriter, req *http.Request, sdkKind basicty
 			}
 			valueObj.Maybe("debugEventsUntilDate", flag.DebugEventsUntilDate != 0).
 				Float64(float64(flag.DebugEventsUntilDate))
```
cwaldren-ld, line 272: All of this to say.. here's the new behavior implemented by the PR.

integrationtests/api_helpers_test.go
```diff
@@ -32,7 +33,8 @@ func (a *apiHelper) logResult(desc string, err error) error {
 		return nil
 	}
 	addInfo := ""
-	if gse, ok := err.(ldapi.GenericOpenAPIError); ok {
+	var gse *ldapi.GenericOpenAPIError
+	if errors.As(err, &gse) {
```
cwaldren-ld, line 37: This little bug (the error is a pointer) was causing response bodies not to be logged.

integrationtests/flag_builder_test.go
```diff
@@ -0,0 +1,131 @@
+//go:build integrationtests
```
cwaldren-ld, line 1: This is a whole new builder scheme just for these tests - we need a simple way to create flags via REST API. The existing methods were single-purpose.

kinyoklion (approved):

PR #458, opened 2024-10-23
feat: add windows binaries

cwaldren-ld:
Customer requests Windows builds of Relay. 

Looks like the test run was successful: https://github.com/launchdarkly/ld-relay/releases/tag/v8.10.0-beta.1

- [x] sanity check run the binary on Windows

BEGIN_COMMIT_OVERRIDE
fix(build): add windows binaries
END_COMMIT_OVERRIDE

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -11,6 +11,7 @@ builds:
   goos: # Goreleaser's default for this is darwin + linux, but let's be explicit
     - darwin
     - linux
+    - windows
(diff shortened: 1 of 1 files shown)


.goreleaser.yml
```diff
@@ -11,6 +11,7 @@ builds:
   goos: # Goreleaser's default for this is darwin + linux, but let's be explicit
```
kinyoklion: Comment could use updated now.

kinyoklion (approved):

PR #480, opened 2025-02-28
feat: Update relay connection metrics to include instance IDs

keelerm84:


keelerm84:
Note that this is against a feature branch. There is additional work to be done before this entire feature is ready to be moved out.

tanderson-ld (approved):

PR #483, opened 2025-03-11
fix: Dedupe event senders by instance ID

keelerm84:
We currently create an event sender for each unique combination of schema version and application tags. With the introduction of instance IDs, we are going to want to report event traffic by this dimension as well.


tanderson-ld (approved):

keelerm84:
Abandoning this as we're going to take a different approach with these events.

PR #491, opened 2025-04-14
feat: Improve relay usage tracking statistics

keelerm84:


internal/metrics/usage_test.go
```diff
@@ -0,0 +1,369 @@
...
+	env := NewEnvironmentMetricUsage("relayID", publisher, 1*time.Hour)
+
+	env.usageActivityMessage(&usageActivityMessage{
+		kind:             UsageActivityKindCount,
+		userAgent:        "userAgent",
+		platformCategory: "platform",
+		instanceID:       "instanceID",
+	})
+
+	time.Sleep(100 * time.Millisecond)
```
tanderson-ld: Any other option besides sleeps?

internal/metrics/usage.go
```diff
@@ -0,0 +1,228 @@
...
+					// Record how far into the metric interval we are so we can
+					// calculate the total stream time later. See #1 in the
+					// flushInterval comment below.
+					usage.streamingOffset += now.Sub(usage.firstActive)
+
+					e.usages[key] = usage
+				case UsageActivityKindStreamDisconnected:
+					usage, exists := e.usages[key]
+					if !exists {
+						continue
```
tanderson-ld: Is this worth logging?

internal/metrics/usage.go
```diff
@@ -0,0 +1,228 @@
...
+// This metric usage is updated when it receives count or streaming messages
+// from the usage middleware. We then track that usage and report it upstream
+// through our event processors.
+type metricUsage struct {
+	firstActive time.Time
+	lastActive  time.Time
+
+	streamingCount      int64
+	streamingOffset     time.Duration
+	streamingRunningSum time.Duration
```
tanderson-ld: Comments on what the offset and running sum represent may help a future developer if they need to tweak the math down below.

internal/metrics/usage.go
```diff
@@ -0,0 +1,228 @@
...
+						usage = &metricUsage{
+							firstActive: now,
+						}
+					}
+					usage.lastActive = now
+					usage.streamingCount += 1
+					// Record how far into the metric interval we are so we can
+					// calculate the total stream time later. See #1 in the
+					// flushInterval comment below.
+					usage.streamingOffset += now.Sub(usage.firstActive)
```
tanderson-ld: Is it possible to eliminate `streamingOffset` if you do `usage.streamingRunningSum -= now.Sub(usage.firstActive)`?  That has better mirror symmetry with the disconnect code.

internal/metrics/usage.go
keelerm84: Done!

internal/metrics/usage.go
keelerm84: Done.

internal/metrics/usage.go
keelerm84: I don't think so since there isn't anything the customer can do about it, and I doubt they are going to report it.

That being said, I am going to change this slightly so that we still record the activity, which is good for usage purposes.

internal/metrics/usage_test.go
keelerm84: I could inject it in I suppose. But this is common throughout our code bases. The tests run in parallel, and so are already quite fast. 

I reduced the durations though to make them faster for you.

tanderson-ld (approved):

PR #493, opened 2025-04-29
feat: support for graceful shutdown of http server in case of SIGTERM signal

mmatecki:
**Requirements**

[x] I have added test coverage for new or changed functionality
[x] I have followed the repository's pull request submission guidelines
[x] I have validated my changes against all supported platform versions (change is platform agnostic)

**Related issues**

No specific issues referenced.

This PR implements graceful HTTP server shutdown functionality in the LD Relay server. The changes include:

* Added graceful shutdown handling for the HTTP server when receiving a SIGTERM signal
* Implemented a 30-second timeout for graceful shutdown
* Added proper error handling for server shutdown scenarios
* Added comprehensive test coverage for the graceful shutdown functionality

The implementation ensures that:

* The server properly handles SIGTERM signals
* In-flight requests are allowed to complete before shutdown
* The shutdown process has a reasonable timeout
* Appropriate logging is provided during the shutdown process

**Additional context**

When we deploy a service in AWS ECS, we use an ALB to manage traffic to the service. The ALB has an "idle timeout" configuration, which controls how long it will keep a connection open when there is no activity.

Despite having the idle timeout configured correctly (we allow for 3 minutes heartbeats as stated in your docs), we notice clusters of 502 Bad Gateway errors at the ALB level. These errors seem to correlate with the timing of deployments, suggesting that the issue occurs when we update or replace the ECS service.

The likely cause is that during a deployment, the ECS service is terminated ungracefully—meaning the old task (container/server) is stopped abruptly before it has a chance to properly close all active connections. This leads to stale connections that the ALB is still holding on to. When the ALB tries to route requests to these outdated, terminated instances, it receives no valid response, resulting in 502 errors.


keelerm84:
Thank you for taking the time to contribute. I don't have time to dig into this myself currently, but a quick test seems to indicate there is still an issue at play.

If I issue a `kill $PID`, I do see the log messages being printed as expected, but the relay itself isn't closing. There must be something hung up internally we need to sort out. As I mentioned, I'm a tad busy for the next couple of iterations, so if you can wait until then, I can try to take a look later. Otherwise, I think we need that sorted.

Also, I like the timeout you added in case it doesn't shutdown in a certain amount of time. To make this production ready for us, I think we would need to add a configuration option for customers to specify that time, maybe with 30 being the default. Is that something you would want to take on?

mmatecki:
Sure - let me take a look at this things 👍

mmatecki:
@keelerm84 I found an issue with that hanging process. I should caught it earlier... 🤦‍♂️ After graceful shutdown I was not closing `errCh`.

Gracefult timeout is now configurable. `configuration.md` is updated. Please take a look whenever you have a while.

Thanks!

mmatecki:
Hey @keelerm84,
I hope you're doing well. Any chance you could spare some time this week to take a look at this PR, please?
TIA!

keelerm84:
Hey @mmatecki, thank you for your patience.

I did some testing this morning, and it shutdowns nicely when there isn't an environment connected. However, if a streaming client is connected, the relay isn't proactively severing that connection. The end result is shutdown now takes 30 seconds for any in-use instance.

I realize that isn't an issue with your PR. Rather, the shutdown method isn't operating as one would expect. But the issue affects existing behavior enough that other customers may be adversely affected in their existing use cases.

Is that something you're interested looking into?

mmatecki:
Hey @keelerm84,

No problem :)

I think that issue actually might be that the connection is long-lived from your side, so it's not going to be completed anytime soon. This graceful timeout makes sense if we have a number of short live request in flight.
So actually in our case it doesn't make sense to wait for connections to close, we just should execute `srv.Shutdown(ctx)` and it will close that connection. Does it sound right to you?

If so, I'll:
* remove graceful timeout
* remove it from config
* update test if required.

mmatecki:
@keelerm84 ignore my previous message, actually I found one more thing we need to do during SIGTERM - we need to call `relay.Close`. Just adding this and additional tests. Will drop you a message in the next couple of hours when it's there.

mmatecki:
@keelerm84 I think I identified final issue and fixed this in the latest commit: https://github.com/launchdarkly/ld-relay/pull/493/commits/f9e3eca40d893f2c3a3cba85e1739aabd12547f7 - it's covered with test. 🤞 we are almost there.

keelerm84 (approved):

keelerm84:
Thank you for your time and continued patience as we worked through these changes. And thank you for this contribution!

mmatecki:
Thank you!

PR #494, opened 2025-05-06
feat: Replace upstream data source with data system

keelerm84:


keelerm84:
Note that this work is against the feat/fdv2 branch, not the current production branches.

internal/datadestination/datadestination.go
```diff
@@ -0,0 +1,102 @@
...
+// SetBasis defines a new basis for the data store. This means the store must
+// be emptied of any existing data before applying the events. This operation should be
+// atomic with respect to any other operations that modify the store.
+//
+// The selector defines the version of the basis.
+//
+// If persist is true, it indicates that the data should be propagated to any connected persistent
+// store.
+func (d *DataDestinationWrapper) SetBasis(events []subsystems.Change, selector subsystems.Selector, persist bool) {
+	allData, err := subsystems.ToStorableItems(events)
```
tanderson-ld, line 67: Is this conversion expected to reside in this location in the long term?  My thinking is that d.updates would eventually support the FDv2 style SetBasis / ApplyDelta.

internal/datadestination/datadestination.go
```diff
@@ -0,0 +1,102 @@
...
+// respect to any other operations that modify the store.
+//
+// The selector defines the new version of the basis.
+//
+// If persist is true, it indicates that the changes should be propagated to any connected persistent
+// store.
+func (d *DataDestinationWrapper) ApplyDelta(events []subsystems.Change, selector subsystems.Selector, persist bool) {
+	allData, err := subsystems.ToStorableItems(events)
+	if err != nil {
+		d.logging.Loggers.Debugf("Failed to convert events to storable items", "error", err)
```
tanderson-ld, line 90: Is this a problematic state?  Seems like this could exist in today's code though, so guessing it is not a new case?

internal/datadestination/datadestination.go
keelerm84: Yes, that is the plan. In fact, I have another PR coming that has this already.

internal/datadestination/datadestination.go
keelerm84: This will be moved in an upcoming PR so this isn't a problem for now.

There is still the possibility something could fail later on in the fdv1 conversion, but we can talk about that in that PR.

tanderson-ld (approved):

PR #498, opened 2025-05-12
feat: Add support for FDv2 server-side streaming

keelerm84:


internal/relayenv/env_context_impl.go
```diff
@@ -797,22 +823,20 @@ func (q envContextStoreQueries) GetAll(kind ldstoretypes.DataKind) ([]ldstoretyp
...
-				if s, ok := keyedItem.Item.Item.(*ldmodel.Segment); ok && s.Unbounded {
-					hasBigSegment = true
-					break
-				}
+	for _, event := range events {
+		if event.Kind == subsystems.SegmentKind {
+			var segment struct {
+				Unbounded bool `json:"unbounded,omitempty"`
+			}
+			if err := json.Unmarshal(event.Object, &segment); err == nil && segment.Unbounded {
```
tanderson-ld, line 837: Does this indicate this could be handled at a better layer?

internal/streams/stream_events.go
```diff
@@ -32,6 +34,66 @@ func (e deferredEvent) Event() string { return e.name }
...
 func (e deferredEvent) Data() string  { return e.result.Get() }
 
+func MakeEventsForSetBasis(changes []subsystems.Change, selector subsystems.Selector) []eventsource.Event {
+	// First create a new streaming protocol builder
+	protocol := ldservicesv2.NewStreamingProtocol()
+
+	// Add the server intent event
+	protocol = protocol.WithIntent(subsystems.ServerIntent{
+		Payload: subsystems.Payload{
+			ID:     "some-id",
```
tanderson-ld: Consider a self documenting name like `relay-spoofed-id` or something like that so if this is seen in a log it shows intentionality.

internal/streams/stream_events.go
```diff
@@ -32,6 +34,66 @@ func (e deferredEvent) Event() string { return e.name }
...
+			Target: selector.Version(),
+			Code:   subsystems.IntentTransferFull,
+			Reason: "cant-catchup",
+		},
+	})
+
+	// Add each change as a put or delete event
```
tanderson-ld:
```suggestion
	// Add each change as a put
```

internal/streams/stream_provider_server_side_flags.go
```diff
@@ -50,29 +58,56 @@ func (s *serverSideFlagsOnlyStreamProvider) Register(
...
+	if err != nil {
 		return
 	}
-	if item.Item == nil {
-		e.server.Publish(e.channels, MakeServerSideFlagsOnlyDeleteEvent(key, item.Version))
-	} else {
-		e.server.Publish(e.channels, MakeServerSideFlagsOnlyPatchEvent(key, item))
+	for _, collection := range allData {
+		if collection.Kind != ldstoreimpl.Features() {
+			continue
```
tanderson-ld, line 102: Is this ok since this is in the "flags only" provider?

tanderson-ld: Discussed on huddle, this provider ignores segments.

internal/relayenv/env_context_impl.go
keelerm84: Yes. I have a note to myself to explore ways of optimizing this away. But we were doing this for fdv1 so at least it isn't any worse.

internal/streams/stream_events.go
keelerm84: Done.

internal/streams/stream_events.go
keelerm84: Removing this whole comment since it's super redundant.

tanderson-ld (approved):

PR #511, opened 2025-06-07
feat: support compression for http responses.

mmatecki:
**Requirements**

[x] I have added test coverage for new or changed functionality
[x] I have followed the repository's pull request submission guidelines
[x] I have validated my changes against all supported platform versions (change is platform agnostic)

**Related issues**

No specific issues referenced.

This PR implements HTTP response compression functionality for the LD Relay Proxy. The changes include:

- Created a new compression middleware supporting both Gzip and Brotli compression algorithms
- Added configuration options for enabling compression and setting compression levels
- Updated documentation with new `[HTTP]` configuration section
- Added dependency on `github.com/andybalholm/brotli` for Brotli compression support

The implementation ensures that:

- Compression is automatically negotiated based on client Accept-Encoding headers
- Brotli compression is preferred over Gzip when both are supported by the client
- Compression levels are validated (Brotli: 1-11, Gzip: 1-9) with default set to 6
- To preserve current behaviour by default compression is disabled.

For Brotli alghorithm `github.com/google/brotli/go/cbrotli` was considered. However that one depends on some C headers and one from `github.com/andybalholm` is pure Golang.

**Additional context**

In our implementation of ldrelayproxy we can observe a lot of data going through ALBs (terabytes per month) - it has impact on the cost of ALBs. Additionally as payloads are sent to customers' browser it has also impact on their experience (more data to download, additional delay before rendering).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,7 @@ ENV GOPATH=/go
 
 RUN go build -a -o ldr .
 
-FROM alpine:3.22.0
+FROM alpine:3.22.1
 
 RUN addgroup -g 1000 -S ldr-user && \
     adduser -u 1000 -S ldr-user -G ldr-user && \
--- a/Dockerfile.goreleaser
+++ b/Dockerfile.goreleaser
@@ -1,6 +1,6 @@
 # This is a Dockerfile used for release (published to dockerhub by goreleaser)
 
-FROM alpine:3.22.0
+FROM alpine:3.22.1
 # See "Runtime platform versions" in CONTRIBUTING.md
 
 RUN apk add --no-cache \


keelerm84:
Hey @mmatecki, my apologies for not getting back to you sooner. I haven't had time to review this PR in detail yet, but I did want to raise the question of whether or not there is an existing middleware we can use to solve this. 

When I run this locally and try to connect to the streaming endpoint, I see the relay generating an error if the `Accept-Encoding` header has been set on the request. Are you experiencing this issue?

mmatecki:
Hey @keelerm84 

Sorry, similarly busy... I could reproduce that issue, and I have fix for that but need to double check everything. I'll push it through the weekend or early next week. Theoretically there's middleware for that: https://github.com/gorilla/handlers/blob/main/compress.go but it supports only gzip. 

Actually if you feel more comfortable with that one we could keep everything from PR related to enabling compression and level of compression for gzip and use that one instead. Brotli compression could be added in future if required.

What do you think? What would be your preference?

mmatecki:
Hey @keelerm84 
Could you please give it a second look. Now it's updated and deals with streaming request. It still have separate compression code to support Brotli and Gzip - but if you prefer to support just gzip with Gorilla's middleware - I can switch to it.

PR #512, opened 2025-06-10
feat: Add server-sdk FDv2 polling endpoint

keelerm84:


relay/relay_endpoints.go
```diff
@@ -146,13 +147,127 @@ func streamHandlerV2(streamProvider streams.StreamProvider, logMessage string) h
...
+								Object:  writer.Bytes(),
+							},
+						})
+					} else {
+						clientCtx.Env.GetLoggers().Error("Error casting keyed item to feature segment")
+						w.WriteHeader(http.StatusInternalServerError)
+						return
+					}
+				default:
+					clientCtx.Env.GetLoggers().Errorf("Unexpected data kind in store snapshot: %s", kind)
```
tanderson-ld, line 246: Consider including in log message that a new entity kind may exist and the relay may need to be updated to support it.  At least I think that is the case that may hit this in the future.

Though perhaps there is not precedent for such a log message in any of our libraries.

relay/relay_endpoints.go
```diff
@@ -146,13 +147,127 @@ func streamHandlerV2(streamProvider streams.StreamProvider, logMessage string) h
...
+	Events []payloadEvent `json:"events"`
+}
+
+type payloadEvent struct {
+	Event     string `json:"event"`
+	EventData any    `json:"data"`
+}
+
+// Server-side SDK polling endpoint: app.ld.com/sdk/poll/
+func pollHandlerV2(w http.ResponseWriter, req *http.Request) {
```
tanderson-ld, line 160: Is this testable?

tanderson-ld (approved):

relay/relay_endpoints.go
keelerm84: Yep, just added one. My next task is going to be shoring up some testing so I plan to add more coverage for the relay in general coming up.

PR #525, opened 2025-07-25
chore: Fix payload transferred encoding for polling response

keelerm84:
The data field in the payload-transferred event was being double-encoded
for polling, resulting in invalid state comparisons between polling and
streaming.


louis-launchdarkly (approved):

tanderson-ld (approved):
Assuming selector has state and version inside it.

PR #532, opened 2025-08-14
fix: change http compression provider

mmatecki:
**Requirements**

- [X] I have added test coverage for new or changed functionality
- [X] I have followed the repository's [pull request submission guidelines](../blob/master/CONTRIBUTING.md#submitting-pull-requests)
- [X] I have validated my changes against all supported platform versions

After deploying HTTP compression to production, we saw excellent compression: HTTP response sizes dropped from ~238 KB to ~48 KB. However, under heavy load, the service’s containers started crashing due to out-of-memory (OOM) kills.

After rolling back and investigating, it appears github.com/gorilla/handlers/CompressionHandler is implemented suboptimally: it doesn’t use a pool and allocates memory for each request, and the garbage collector (GC) can’t keep up with the churn.

I looked for alternatives and found [klauspost/compress](https://github.com/klauspost/compress); its gzhttp package uses pools and is more sophisticated (for example doesn't compress content with size lower than 1kb).

I switched implementations and ran a few tests against both. I varied concurrency and total request count, and recorded peak memory usage. The table below shows the results (average peak memory across three runs):

image

The difference is especially visible at higher concurrency levels.

Measurements were taken using `TestLoadCompression`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -60,6 +60,7 @@ require (
 
 require (
 	github.com/alecthomas/units v0.0.0-20231202071711-9a357b53e9c9
+	github.com/klauspost/compress v1.18.0
 	github.com/launchdarkly/api-client-go/v13 v13.0.1-0.20230420175109-f5469391a13e
 )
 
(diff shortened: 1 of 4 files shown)


mmatecki:
@keelerm84 sorry but it's not end of story with compression... Could you take a look, please? Thank you!

keelerm84 (approved):

PR #538, opened 2025-10-09
feat: Introduce shared object cache to reduce retained memory

keelerm84:
This is going to be a cut as a pre-release first for testing which is why I'm merging it into a pre-release style branch.


jsonbailey (approved):

keelerm84:
bugbot review

PR #558, opened 2025-12-02
fix: Remove double slash in diagnostic url debug message

jsonbailey:
> [!NOTE]
> Fixes diagnostic event proxy debug log to concatenate base URI and path without inserting an extra slash.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f9afd95ab6dbc57a1f9b212bbaf0b080dec67b77. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/internal/events/event-relay.go
+++ b/internal/events/event-relay.go
@@ -126,7 +126,7 @@ func (d *diagnosticEventEndpointDispatcher) dispatch(w http.ResponseWriter, req 
 	consumeEvents(w, req, d.loggers, func(body []byte) {
 		// We are just operating as a reverse proxy and passing the request on verbatim to LD; we do not
 		// need to parse the JSON.
-		d.loggers.Debugf("Received diagnostic event to be proxied to %s/%s", d.baseURI, d.uriPath)
+		d.loggers.Debugf("Received diagnostic event to be proxied to %s%s", d.baseURI, d.uriPath)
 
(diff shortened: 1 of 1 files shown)


joker23 (approved):

internal/events/event-relay.go
```diff
@@ -126,7 +126,7 @@ func (d *diagnosticEventEndpointDispatcher) dispatch(w http.ResponseWriter, req
 	consumeEvents(w, req, d.loggers, func(body []byte) {
 		// We are just operating as a reverse proxy and passing the request on verbatim to LD; we do not
 		// need to parse the JSON.
-		d.loggers.Debugf("Received diagnostic event to be proxied to %s/%s", d.baseURI, d.uriPath)
+		d.loggers.Debugf("Received diagnostic event to be proxied to %s%s", d.baseURI, d.uriPath)
```
keelerm84, line 129: Have we previously ensured that the d.baseURI always has a trailing slash somewhere? Or that d.uriPath has a leading slash?

jsonbailey: From what I saw, all the instances did have the leading slash. But it is on us to ensure that it always does using this method. We could construct the url to ensure it is always properly built, but this is also for a debug message so I'm not sure how critical it actually is.

keelerm84 (approved):

PR #583, opened 2026-03-03
feat!: Switch opencensus for opentelemetry

keelerm84:
> [!NOTE]
> **High Risk**
> High risk because it removes the existing Datadog/Stackdriver/Prometheus OpenCensus exporters and replaces the metrics/tracing pipeline with OpenTelemetry OTLP, changing runtime telemetry behavior and configuration surface area.
> 
> **Overview**
> **Switches Relay metrics from OpenCensus to OpenTelemetry.** The PR removes the OpenCensus exporter framework (Datadog/Stackdriver/Prometheus) and rewrites `internal/metrics` to record metrics via OTel instruments and optionally export via OTLP (gRPC/HTTP).
> 
> **Configuration and docs are updated accordingly.** `MetricsConfig`/Prometheus port defaults are removed, a new `[OpenTelemetry]`/env-var driven `OpenTelemetryConfig` is added with protocol validation, and `docs/metrics.md` is rewritten around OTLP push to Prometheus/Collector. CI drops the Prometheus endpoint test, and dependencies are updated to add OTel/OTLP libraries while removing OpenCensus-related modules.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b1074ca69aa0bc81e3488fc4a06eb95153cfe5eb. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


config/config_validation.go
```diff
@@ -273,6 +276,15 @@ func validateConfigDatabases(result *ct.ValidationResult, c *Config, loggers ldl
 	}
 }
 
+func validateConfigMetrics(result *ct.ValidationResult, c *Config) {
+	if c.OpenTelemetry.Enabled {
+		protocol := strings.ToLower(c.OpenTelemetry.Protocol)
+		if protocol != "" && protocol != "grpc" && protocol != "http" {
+			result.AddError(nil, errOTLPInvalidProtocol)
+		}
+	}
```
tanderson-ld, line 285: Consider validating even if disabled as it may be a chance to educate/correct an improper config.

internal/metrics/measures.go
```diff
@@ -3,120 +3,146 @@ package metrics
...
-		logging.GetGlobalContextLoggers(ctx).Errorf(`Failed to create tags: %s`, err)
-	} else {
-		for _, m := range measure.measures {
-			ctx, _ := tag.New(ctx, measure.tags...)
-			stats.Record(ctx, m.M(1))
-			defer stats.Record(ctx, m.M(-1))
-		}
+	if em.collector != nil {
+		em.collector.RecordConnectionChange(measure.platformCategory, sanitizedUA, sanitizedWrapper, 1)
+		defer em.collector.RecordConnectionChange(measure.platformCategory, sanitizedUA, sanitizedWrapper, -1)
```
tanderson-ld, line 101: Is it ok that the order of defers is same as order of instrument/collector calls?  Or should it be FILO?  If they are independent, doesn't seem to matter.

internal/metrics/measures.go
```diff
@@ -3,120 +3,146 @@ package metrics
...
+		requests:       requests,
+	}, nil
 }
 
-func makeBrowserTags() []tag.Mutator {
-	return []tag.Mutator{tag.Insert(platformCategoryTagKey, browserTagValue)}
-}
+// WithGauge increments the specified metric before running the function and then decrements it (for use with
+// the active connection metrics).
+func WithGauge(em *EnvironmentManager, instruments *Instruments, userAgent, sdkWrapper string, f func(), measure Measure) {
```
tanderson-ld, line 84: Consider making name `WithGauge` connection specific

internal/metrics/metrics.go
```diff
@@ -58,44 +53,62 @@ type shutdown struct {
...
 ) (*Manager, error) {
 	metricsRelayID := uuid.New()
 
-	exporters, err := registerExporters(allExporterTypes(), metricsConfig, loggers)
-	if err != nil { // COVERAGE: can't make this happen in unit tests
-		return nil, err
-	}
+	res := resource.NewWithAttributes(
+		semconv.SchemaURL,
+		semconv.ServiceName("ld-relay"),
```
tanderson-ld: Should the service name be flavored with information provided by the customer? (there is another spot with "ld-relay" string, same question)

internal/metrics/otlp.go
```diff
@@ -0,0 +1,86 @@
...
+	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
+)
+
+func newOTLPExporters(
+	otlpConfig config.OpenTelemetryConfig,
+	loggers ldlog.Loggers,
+) ([]sdkmetric.Option, error) {
+	protocol := strings.ToLower(otlpConfig.Protocol)
+	if protocol == "" {
+		protocol = "grpc"
```
tanderson-ld, line 22: Why grpc vs http for default?  Just curious.

internal/metrics/otlp.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	opts := []sdkmetric.Option{sdkmetric.WithReader(sdkmetric.NewPeriodicReader(metricExporter))}
+
+	loggers.Infof("Successfully registered OTLP metrics exporter (protocol=%s, endpoint=%s)", protocol, otlpConfig.Endpoint)
+
+	return opts, nil
+}
+
+func parseHeaders(headers string) map[string]string {
```
tanderson-ld, line 51: Any common header parsing utils? Or standard lib?

internal/metrics/relay_metrics_collector.go
```diff
@@ -0,0 +1,217 @@
...
+		loggers:            loggers,
+		closer:             make(chan struct{}),
+		intervalStartTime:  time.Now(),
+		pollingDataIsDirty: false,
+		currentConnections: make(map[connectionsKeyType]int64),
+		newConnections:     make(map[connectionsKeyType]int64),
+		pollingCounts:      make(map[connectionsKeyType]pollingCounts),
+	}
+
+	flushTicker := time.NewTicker(flushInterval)
```
tanderson-ld, line 91: Is this relay metrics collector a rewrite of another piece of code and is still necessary or is it new functionality?  I'm currently assuming new functionality and so Otel related.

Maybe I am misunderstanding the structure.  In other SDKs that have used Otel, we haven't needed to implement periodic flushing of Otel data and used existing open source otel components.  Perhaps you don't want the dependency?

internal/metrics/relay_metrics_collector.go
```diff
@@ -0,0 +1,217 @@
...
+	if c.pollingDataIsDirty {
+		for k, v := range c.pollingCounts {
+			if v.running != v.lastReported {
+				event.PollingCounts = append(event.PollingCounts, pollingMetric{
+					UserAgent:        k.userAgent,
+					SDKWrapper:       k.sdkWrapper,
+					PlatformCategory: k.platformCategory,
+					Count:            v.running - v.lastReported,
+				})
+				v.lastReported = v.running
```
tanderson-ld, line 183: Is it ok to change this if a publish error occurs?  Since the pipeline can't guarantee delivery, I am guessing this is fine.

internal/sharedtest/metrics.go
```diff
@@ -1,122 +1,79 @@
...
-		case d := <-e.dataCh:
-			loggers.Infof("exporter got metrics: %+v", d)
-			if fn(d) {
-				return
-			}
 		case <-deadline:
-			require.Fail(t, "timed out waiting for metrics data")
+			require.Fail(t, "timed out waiting for span data")
+			return tracetest.SpanStub{}
+		case <-time.After(time.Millisecond * 10):
```
tanderson-ld: Magic test duration

internal/metrics/measures.go
keelerm84: They are two independent systems.

Originally, we tracked these metrics via opencensus. Customers could use that data, but we also used it ourselves as a way to gather usage statistics. But tying those two together isn't ideal.

So in the new mode, we record the change for opentelemetry purposes, but we also record it for our internal statistic reporting purposes.

internal/metrics/otlp.go
keelerm84: Claude made this choice. I don't feel terribly strongly about it, other than grpc is a little more performant so it seems like a reasonable default.

internal/metrics/otlp.go
keelerm84: The standard library approach would need us to modify the string before wrapping it in a reader and passing it through some parsing. I think this is small enough a simple header parsing utility is fine.

internal/metrics/relay_metrics_collector.go
keelerm84: The metrics collector is for gathering usage statistics and reporting it through our internal relayMetrics events. This doesn't have anything to do with otel.

The reason it's part of this PR is because it used to rely on opencensus to work. This commit breaks those two systems apart.

internal/metrics/relay_metrics_collector.go
keelerm84: It's at feature parity with how it behaved previously.

Ideally, we would have better handling for when events fail to send, but I think that's outside the scope of this PR.

internal/sharedtest/metrics.go
keelerm84: This is dead code I need to remove.

internal/metrics/measures.go
keelerm84: I'm going to hold off on renaming these until the next few tasks when I start adding some more metrics. Then I'll have a better feel for how I want to name these.

internal/metrics/metrics.go
keelerm84: Done!

tanderson-ld (approved):

PR #593, opened 2026-03-16
feat: cache autoconfig in store, init from store

LD-Sfalzon:
**Requirements**

- [ ] I have added test coverage for new or changed functionality
- [ X] I have followed the repository's [pull request submission guidelines](../CONTRIBUTING.md#submitting-pull-requests)
- [ X] I have validated my changes against all supported platform versions

**Related issues**
if a relay proxy is unable to connec to LD Streaming service to retrieve auto config the persistent cache does not provide a "resileincy" and it cannot be used for auto scaling during a LD incident

auto config can now be saved to the persistent cache, encrypted by default with the auto config key (configurable) 
I have tested this with Valkey only 

**Additional context**

---

> [!NOTE]
> **Medium Risk**
> Touches AutoConfig stream lifecycle and adds new encrypted persistence to Redis/DynamoDB, so bugs could affect startup/availability or cause stale/failed cache reads/writes.
> 
> **Overview**
> Adds an **optional persistent AutoConfig cache** (Redis/Valkey hash or DynamoDB partition) that stores environments/filters and is used to *bootstrap Relay on startup* when `AUTO_CONFIG_CACHE_KEY` is set.
> 
> `StreamManager` now races an async cache read against the SSE connection, applies cached `put` data if it arrives first, and persists authoritative stream updates back to the cache on `put/patch/delete` events; shutdown now waits for the subscribe goroutine to exit before closing the cache.
> 
> Introduces `internal/autoconfigcache` with AES-256-GCM encryption (HKDF-derived key from `AUTO_CONFIG_CACHE_ENCRYPTION_KEY` or the AutoConfig key), adds config/docs + validation for the new cache settings, and wires the store into Relay initialization.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ca7383f608db414eddd29bbf36db8a9914184377. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/autoconfigcache/dynamodb_store.go
```diff
@@ -0,0 +1,303 @@
...
+}
+
+func (s *dynamoDBStore) Upsert(ctx context.Context, kind autoconfig.CacheKind, id string, data interface{}) error {
+	sortKey := cacheField(kind, id)
+	item, err := s.buildItem(sortKey, data)
+	if err != nil {
+		return fmt.Errorf("AutoConfig cache: failed to build item %q: %w", sortKey, err)
+	}
+	if !s.checkSizeLimit(item, sortKey) {
+		return nil
```
kinyoklion, line 300: Probably should be a log here.

config/config.go
kinyoklion, line 1: If a cache-key exist, then that implies we want to use it for initialization, instead of having two settings.

internal/autoconfigcache/dynamodb_store.go
keelerm84: Turns out there is a lot. It's just in the check function.

config/config.go
keelerm84: Done.

internal/autoconfigcache/dynamodb_store.go
```diff
@@ -0,0 +1,277 @@
...
+			RequestItems: map[string][]types.WriteRequest{
+				s.table: batch,
+			},
+		})
+		if err != nil {
+			return fmt.Errorf("AutoConfig cache DynamoDB batch write: %w", err)
+		}
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
keelerm84: It isn't silently dropped as we issue a log statement. Continuing on failure in this case is better as this is serving solely as a remediation effort, and having some environments is better than no environments.

internal/autoconfigcache/crypto.go
```diff
@@ -0,0 +1,57 @@
+package autoconfigcache
```
keelerm84, line 14: @launchdarkly/team-product-security mind taking a look at this new addition and let me know if it seems acceptable?

LD-Sfalzon:
@keelerm84  sorry there are a few changes which I have made which you will probably need to review

internal/relayenv/env_context_impl.go
```diff
@@ -486,8 +527,64 @@ func (c *envContextImpl) removeCredential(oldCredential credential.SDKCredential
...
+	c.globalLoggers.Warnf(
+		"LaunchDarkly unavailable for %q during SDK init (%v); using persistent store only for flags until connectivity returns",
+		name, err,
+	)
+	c.mu.Lock()
+	c.sdkConfig.DataSource = ldcomponents.ExternalUpdatesOnly()
+	c.featureFlagUpdateSource = "daemon"
+	c.mu.Unlock()
+
+	client2, err2 := c.sdkClientFactory(sdkKey, c.sdkConfig, c.sdkInitTimeout)
```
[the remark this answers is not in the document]
LD-Sfalzon: @keelerm84  do you think this is actually an issue? i don't think it is because the key rotation cant happen while its offline and once it comes online it doesn't revert back to the persistent store as the DataSource

internal/relayenv/env_context_impl.go
```diff
@@ -532,6 +629,93 @@ func (c *envContextImpl) startSDKClient(sdkKey config.SDKKey, readyCh chan<- Env
 	}
 }
 
+func (c *envContextImpl) startStreamingRestoreMonitor() {
+	if c.streamingRestoreStop == nil {
+		return
+	}
+	go c.runStreamingRestoreMonitor()
+}
```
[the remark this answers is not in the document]
LD-Sfalzon: @keelerm84  this one could be an issue maybe ? 


I am testing it, i rotated a key  and i observe two clients are now running

2026/04/13 16:07:55.878272 INFO: AutoConfiguration Received configuration for 1 environment(s)
2026/04/13 16:07:55.878319 INFO: Finished processing auto-configuration data
2026/04/13 16:07:56.421403 INFO: RedisDataStore: Initialized with 772 items
2026/04/13 16:07:56.421991 INFO: [env: ...0ca5] LaunchDarkly streaming is active
2026/04/13 16:07:56.422013 INFO: [env: ...0ca5] Initialized LaunchDarkly client
2026/04/13 16:07:56.422043 INFO: Initialized LaunchDarkly client for "test Test" (SDK key ...bd2e)
2026/04/13 16:07:56.591881 INFO: RedisDataStore: Initialized with 772 items
2026/04/13 16:07:56.592117 INFO: [env: ...0ca5] LaunchDarkly streaming is active
2026/04/13 16:07:56.592126 INFO: [env: ...0ca5] Initialized LaunchDarkly client
2026/04/13 16:07:56.592139 INFO: Initialized LaunchDarkly client for "test Test" (SDK key ...4e7d)

ill keep the relay offline while there are two valid sdk keys and start it back up and observe what happens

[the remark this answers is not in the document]
LD-Sfalzon: i tested this scenario the expiry worked as expected

config/auto_degrade.go
```diff
@@ -0,0 +1,17 @@
+package config
+
+import "strings"
+
+// AutoDegradeSDKWhenLDUnavailable is true when the embedded LaunchDarkly SDK may temporarily use
+// ExternalUpdatesOnly (read flags only from the persistent store) if it cannot reach LaunchDarkly
+// during initialization. Requires auto-config with a persistent cache key and a persistent flag
+// data store. Permanent [Main] DaemonMode disables this (daemon-only behavior is explicit).
+func (c Config) AutoDegradeSDKWhenLDUnavailable() bool {
```
pkaeding: Would it make sense to split the autodegrade change into a separate PR from the RAC caching?

pkaeding:
Could we add test coverage for the autoconfigcache package, and document the threat model that we are trying to protect against with encrypting this?

internal/relayenv/env_context_impl.go
```diff
@@ -532,6 +629,93 @@ func (c *envContextImpl) startSDKClient(sdkKey config.SDKKey, readyCh chan<- Env
...
+func (c *envContextImpl) swapFromDaemonToStreaming() error {
+	c.mu.RLock()
+	if c.closed {
+		c.mu.RUnlock()
+		return fmt.Errorf("environment is closed")
+	}
+	c.mu.RUnlock()
+
+	sdkKey := c.keyRotator.SDKKey()
+	c.mu.Lock()
```
pkaeding: Need to re-check c.closed after acquiring the lock (other places too)

config/auto_degrade.go
LD-Sfalzon: I had the same thought but the RAC caching doesn't really serve a purpose if the result is a valid config which does not accept connections becuase it wont  populate flag rules from the persistent cache

keelerm84: > I had the same thought but the RAC caching doesn't really serve a purpose if the result is a valid config which does not accept connections becuase it wont populate flag rules from the persistent cache

Except it 100% will serve flag values. You just have to wait until the init timeout and you have to have configured the relay to ignore initialization errors. 

This definitely needs to be a separate PR, though the feature feels redundant with the configuration options we are already presenting.

internal/relayenv/env_context_impl.go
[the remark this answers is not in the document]
keelerm84: It may not be a problem in practice, but we should still treat objects under a mutex as always requiring some level of lock access when interacting with them. Assumptions might change later on, and we want to be able to rely on tooling to identify those types of those.

config/auto_degrade.go
LD-Sfalzon: Sorry I didnt realise I had a stupid long timeout in my config from a previous test

LD-Sfalzon:
@pkaeding  in relation to this question

> Could we add test coverage for the autoconfigcache package, and document the threat model that we are trying to protect against with encrypting this?

I have added a document in relation to the threat model this protects against

docs/autoconfig-cache-threat-model.md
```diff
@@ -0,0 +1,42 @@
...
+- **Accidental leakage**: logs, support bundles, or screen shares that include stored payloads.
+
+## Security goals
+
+| Goal | Mechanism |
+|------|-----------|
+| **Confidentiality at rest** | Each cached item is encrypted with **AES-256-GCM** before write. The encryption key is derived with **SHA-256** from `cacheEncryptionKey`, or from the AutoConfig key if `cacheEncryptionKey` is unset (see [configuration](./configuration.md)). |
+| **Integrity of cached blobs** | GCM authentication rejects ciphertext that was truncated or tampered with in the store. |
+
+## Out of scope (non-goals)
```
keelerm84: I think you could remove from here down. Probably more information than we need to include for public documentation.

 I had also pushed a small comment to the crypto.go file which I think is probably sufficient for what Patrick wanted. I'm actually fine removing this file entirely.

LD-Sfalzon: lets remove it entirely then 
less things to maintain

internal/autoconfigcache/crypto.go
```diff
@@ -0,0 +1,70 @@
...
+	"errors"
+	"io"
+	"strings"
+
+	"github.com/launchdarkly/ld-relay/v8/config"
+)
+
+// resolveEncryptionKey returns a 32-byte key for AES-256 by deriving it via SHA-256.
+// If CacheEncryptionKey is set, it is used as the input. Otherwise the AutoConfig key is used.
+func resolveEncryptionKey(c config.Config) []byte {
```
pkaeding, line 33: It would be better to use a proper KDF, eg https://pkg.go.dev/golang.org/x/crypto/hkdf

keelerm84: done.

docs/configuration.md
```diff
@@ -85,16 +85,20 @@ to authorize using that credential. Relay will periodically check for expired cr
...
+|--------------------------|---------------------------------------|:-------:|:--------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `key`                    | `AUTO_CONFIG_KEY`                     | String  |         | A valid Relay Proxy automatic configuration key.                                                                                                                                                                                    |
+| `cacheKey`               | `AUTO_CONFIG_CACHE_KEY`                | String  |         | Enables AutoConfig caching in the persistent store (Redis/Valkey or DynamoDB). Requires Redis or DynamoDB. Multiple Relay instances can share the same store by using different cache keys. _(7)_ |
+| `cacheEncryptionKey`     | `AUTO_CONFIG_CACHE_ENCRYPTION_KEY`    | String  | _(auto)_ | Encrypts cached AutoConfig data. Can be any string; derived to a 256-bit AES key via SHA-256. If omitted, the AutoConfig key is used. _(7)_ |
```
pkaeding:
```suggestion
| `cacheEncryptionKey`     | `AUTO_CONFIG_CACHE_ENCRYPTION_KEY`    | String  | _(auto)_ | Encrypts cached AutoConfig data. must be a high-entropy secret  (e.g., from openssl rand -base64 32 or a password manager). This value is not processed through a password-hardening KDF. _(7)_ |
```

internal/autoconfigcache/crypto_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+	_, err = decrypt(ciphertext, key2)
+	assert.Error(t, err)
+}
+
+func TestDecryptFailsWithTruncatedCiphertext(t *testing.T) {
+	key := deriveKey([]byte("key"))
+	_, err := decrypt([]byte("short"), key)
+	assert.Error(t, err)
+}
```
pkaeding, line 113: The existing truncation test is good, but it doesn't verify that GCM's authentication tag is actually being checked — a bug that dropped the tag or used `cipher.NewCipher` directly (CTR mode, no auth) would slip through. Suggest expanding this area to cover tamper detection across the nonce, body, and tag regions.

```suggestion
func TestDecryptFailsWithTruncatedCiphertext(t *testing.T) {
	key := deriveKey([]byte("key"))
	_, err := decrypt([]byte("short"), key)
	assert.Error(t, err)
}

func TestDecryptRejectsTamperedCiphertext(t *testing.T) {
	key := deriveKey([]byte("test-key"))
	plaintext := []byte("sensitive environment data including SDK keys")

	ciphertext, err := encrypt(plaintext, key)
	require.NoError(t, err)

	// Ciphertext layout: nonce (12 bytes) || encrypted body || GCM tag (16 bytes).
	// Flipping a bit anywhere in that blob must cause Open() to fail.
	const nonceSize = 12
	const tagSize = 16
	bodyStart := nonceSize
	bodyEnd := len(ciphertext) - tagSize
	require.Greater(t, bodyEnd, bodyStart, "ciphertext should have a non-empty body")

	cases := []struct {
		name  string
		index int
	}{
		{"flip bit in nonce", 0},
		{"flip bit in nonce (last byte)", nonceSize - 1},
		{"flip bit in encrypted body", bodyStart},
		{"flip bit in middle of body", bodyStart + (bodyEnd-bodyStart)/2},
		{"flip bit in GCM auth tag", bodyEnd},
		{"flip bit in last byte of tag", len(ciphertext) - 1},
	}

	for _, tc := range cases {
		t.Run(tc.name, func(t *testing.T) {
			tampered := append([]byte(nil), ciphertext...)
			tampered[tc.index] ^= 0x01

			_, err := decrypt(tampered, key)
			assert.Error(t, err, "decrypt must reject tampered ciphertext at index %d", tc.index)
		})
	}
}

func TestDecryptRejectsAppendedBytes(t *testing.T) {
	key := deriveKey([]byte("test-key"))
	ciphertext, err := encrypt([]byte("payload"), key)
	require.NoError(t, err)

	tampered := append(append([]byte(nil), ciphertext...), 0x00, 0x01, 0x02)
	_, err = decrypt(tampered, key)
	assert.Error(t, err)
}

func TestDecryptRejectsTruncatedTag(t *testing.T) {
	key := deriveKey([]byte("test-key"))
	ciphertext, err := encrypt([]byte("payload"), key)
	require.NoError(t, err)

	// Lop off the last byte of the GCM tag — auth must fail.
	tampered := ciphertext[:len(ciphertext)-1]
	_, err = decrypt(tampered, key)
	assert.Error(t, err)
}
```

Note on the magic numbers: `nonceSize = 12` and `tagSize = 16` are the AES-GCM defaults and are stable parts of the algorithm spec, so hardcoding them keeps the test readable. If you'd prefer to avoid the literals, the test can instead build its own `cipher.NewGCM` and read `gcm.NonceSize()` / `gcm.Overhead()`, at the cost of a bit more setup.

via LD Research 🤖

pkaeding (approved):
I suggested a couple more tests around the crypto stuff, to try to ensure that there are no bugs there. but otherwise, LGTM!

docs/configuration.md
```diff
@@ -85,16 +85,20 @@ to authorize using that credential. Relay will periodically check for expired cr
...
+| `key`                    | `AUTO_CONFIG_KEY`                     | String  |         | A valid Relay Proxy automatic configuration key.                                                                                                                                                                                    |
+| `cacheKey`               | `AUTO_CONFIG_CACHE_KEY`                | String  |         | Enables AutoConfig caching in the persistent store (Redis/Valkey or DynamoDB). Requires Redis or DynamoDB. Multiple Relay instances can share the same store by using different cache keys. _(7)_ |
+| `cacheEncryptionKey`     | `AUTO_CONFIG_CACHE_ENCRYPTION_KEY`    | String  | _(auto)_ | Encrypts cached AutoConfig data. must be a high-entropy secret  (e.g., from openssl rand -base64 32 or a password manager). This value is not processed through a password-hardening KDF. _(7)_ |
```
kinyoklion, line 92: I am not sure about saying KDF in the documentation for this key without some kind of clarification.

internal/autoconfigcache/crypto.go
```diff
@@ -0,0 +1,81 @@
+// The cache stores environment metadata including SDK keys, mobile keys, and project
+// configuration. These credentials could be used to connect to LaunchDarkly or receive
+// flag data if exposed.
+//
+// The persistent store (Redis or DynamoDB) should already be access-controlled, but
```
kinyoklion: Is this too detailed, and too much like a guarantee?

internal/autoconfig/stream_manager.go
```diff
@@ -103,24 +139,79 @@ func NewStreamManager(
...
+		s.subscribe(readyCh)
+	}()
 	return readyCh
 }
 
-// Close permanently shuts down the stream.
+// Close permanently shuts down the stream and waits for the subscribe goroutine
+// to exit before closing the cache, ensuring no in-flight cache writes are interrupted.
+// Safe to call even if Start() was never called.
 func (s *StreamManager) Close() {
```
kinyoklion, line 178: Can close be called while start is still in progress? (I assume at least they cannot be Close and then Start.)

internal/autoconfig/stream_manager.go
```diff
@@ -103,24 +139,79 @@ func NewStreamManager(
...
+// The cache read and stream connection are started concurrently. If the cache returns first,
+// its data is applied so Relay can serve immediately. If the stream's first PUT arrives first,
+// the cache read is cancelled and its result discarded.
 func (s *StreamManager) Start() <-chan error {
+	// Start the cache read concurrently with the stream connection.
+	cacheCtx, cacheCancel := context.WithCancel(context.Background())
+	cacheCh := make(chan *PutContent, 1)
+	go func() {
+		defer close(cacheCh)
+		content, err := s.cache.GetAll(cacheCtx)
```
kinyoklion, line 152: I assume if the cache is closed while this is executing, then the consequence would basically be the same as cancelling it? The close and the context actually cancelling the quest will be effectively raced. I think it is fine regardless.

internal/autoconfig/stream_manager.go
```diff
@@ -156,22 +247,55 @@ func (s *StreamManager) subscribe(readyCh chan<- error) {
...
+
+		case result := <-streamCh:
+			if result.err != nil {
+				s.loggers.Errorf(logMsgStreamOtherError, result.err)
+				signalReady(result.err)
+				return
+			}
+			stream = result.stream
+
+		case <-s.halt:
```
kinyoklion, line 291: If this happens first then stream results will get buffered into a channel that isn't drained? Maybe?

internal/autoconfigcache/dynamodb_store.go
```diff
@@ -0,0 +1,326 @@
...
+
+func (s *dynamoDBStore) batchWrite(ctx context.Context, requests []types.WriteRequest) {
+	for i := 0; i < len(requests); i += dynamoDBMaxBatchSize {
+		end := i + dynamoDBMaxBatchSize
+		if end > len(requests) {
+			end = len(requests)
+		}
+		batch := requests[i:end]
+		_, err := s.client.BatchWriteItem(ctx, &dynamodb.BatchWriteItemInput{
```
kinyoklion: Does BatchWriteItem return an error in the case where it populates unprocessed keys?

internal/autoconfig/stream_manager.go
keelerm84: Yes, it should essentially behave the same, with some slight difference in logging. We won't mention that the cache read failed if the context was cancelled.

internal/autoconfig/stream_manager.go
keelerm84: In practice, the only way this would occur would be if the relay started up and then immediately received a kill signal.

docs/configuration.md
keelerm84: What do you think about just dropping the last sentence?

internal/autoconfigcache/crypto.go
keelerm84: Updated.

internal/autoconfigcache/dynamodb_store.go
keelerm84: No, it looks like that is handled differently. I have updated the code to deal with that.

internal/autoconfig/stream_manager.go
keelerm84: Updated to deal with this.

kinyoklion (approved):

PR #617, opened 2026-03-31
feat: Add internal event processing OTEL metrics

keelerm84:
Establish an EventMetrics interface pattern for reporting internal
event processing telemetry via OpenTelemetry. Each subsystem defines
its own interface; the metrics package provides an OTEL-backed
implementation (EventMetricsRecorder) that satisfies both the ld-relay
and go-sdk-events interfaces via Go structural typing.

New metrics under the events.received namespace:
- events.received.bytes: bytes of event data received from SDKs
  (renamed from events.ingested.bytes)

New metrics under the events.sent namespace:
- events.sent.count: events successfully delivered to LaunchDarkly
- events.sent.bytes: payload bytes delivered (pre-compression)
- events.sent.failures: events in batches that failed after retries
- events.sent.dropped: events discarded due to capacity overflow
- events.sent.pending: current number of events buffered (gauge)

All metrics are recorded from both the verbatim relay
(HTTPEventPublisher) and the summarizing relay (go-sdk-events
DefaultEventProcessor) paths. A temporary go.mod replace directive
points to the local go-sdk-events worktree for development.

---

> [!NOTE]
> **Medium Risk**
> Adds new OpenTelemetry instruments and wires them into the async event forwarding path (verbatim + summarizing), which could affect event pipeline performance/behavior if mis-instrumented, but is largely additive and guarded with no-op defaults.
> 
> **Overview**
> Adds **internal event-processing telemetry** via a new `events.EventMetrics` interface, with an OTEL-backed `EventMetricsRecorder` that tracks event queue depth, dropped events, bytes sent, successful sends, and failed sends (including `statusCode`).
> 
> Renames the existing ingress metric from `launchdarkly.relay.events.ingested.bytes` to `launchdarkly.relay.events.received.bytes`, updates middleware/docs/tests accordingly, and threads the metrics recorder into both the verbatim (`HTTPEventPublisher`) and summarizing (`go-sdk-events` processor) relay paths.
> 
> Also updates CI to grant `integration-test` workflow `id-token`/`contents` permissions and bumps `go-sdk-events` to `v3.6.0` alongside related module tidy changes.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 26f084d44c5e7a71cce6139a2668de2f517141cd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


internal/events/event_metrics.go
```diff
@@ -0,0 +1,39 @@
...
+	RecordEventsFailedSend(count int, metadata EventSendFailureMetadata)
+
+	// RecordEventsBytesSent is called when a batch of events has been successfully delivered.
+	// The bytes parameter is the size of the serialized event payload before compression.
+	RecordEventsBytesSent(bytes int)
+
+	// RecordPendingEvents is called after any operation that changes the number of events
+	// buffered awaiting delivery. The count parameter is the current total number of events pending.
+	RecordPendingEvents(count int)
+}
```
aaron-zeisler, line 39: Thinking out loud: Would the underlying metrics collection system allow us to "tag" or "label" these metrics with additional metadata? I'm thinking about pieces of data that would help us categorize and analyze the data. Off the top of my head, the version or relay proxy may be a good tag, allowing us to compare different versions. Some customer identifier is most likely a poor choice for a tag, since its cardinality would be very high, but it would allow us to dig into customer-specific issues. Maybe there are other pieces of data that are valuable for slicing and dicing the metric data in different ways.

aaron-zeisler: Ah, I see it now. `EventMetricsRecorder.envKVs` is parsed and transformed into `attributes`. envKVs comes from the `EnvironmentManager`. Do you think there would be many call-specific pieces of data that we'd use as an attribute?

keelerm84: I added that for the events that fail to send. They are including the status code of the failed request. I didn't have any good ideas for the other ones, but if you have some specific ideas, I'm happy to include them.

internal/events/event_publisher.go
```diff
@@ -257,10 +272,24 @@ func (p *HTTPEventPublisher) append(batch eventBatch) {
 			p.overflowed = true
 		}
 		taken = available
+		if dropped := len(batch.events) - taken; dropped > 0 && p.eventMetrics != nil {
```
tanderson-ld: Do you want to record dropped of 0?  Seems like a valid data point.

internal/metrics/measures.go
```diff
@@ -177,3 +210,58 @@ func RecordRequestDuration(ctx context.Context, instruments *Instruments, em *En
...
+	if r.instruments == nil || bytes <= 0 {
+		return
+	}
+	r.instruments.eventsBytesSent.Add(context.Background(), int64(bytes), metric.WithAttributeSet(r.envAttrs))
+}
+
+// RecordEventsFailedSend records the number of events that could not be delivered after all retries.
+// The status code from the metadata is included as an attribute on the metric.
+func (r *EventMetricsRecorder) RecordEventsFailedSend(count int, metadata ldevents.EventSendFailureMetadata) {
+	if r.instruments == nil || count <= 0 {
```
tanderson-ld, line 259: 0 seems like a valid count.  It provides confirmation there are no failures.  Perhaps this is a cost saving measure to not send 0?

internal/events/event_publisher.go
```diff
@@ -351,12 +381,25 @@ func (p *HTTPEventPublisher) flush() {
 				EnableCompression: true,
 			}
 			result := ldevents.SendEventDataWithRetry(sendConfig, ldevents.AnalyticsEventDataKind, p.uriPath, payload, count)
+			if eventMetrics != nil {
```
tanderson-ld: Consider a no-op implementation pattern to eliminate all the various null checks throughout.

tanderson-ld (approved):

internal/metrics/measures.go
keelerm84: Failed events are also counter metrics, so 0 isn't a useful value. Avoiding it prevents the unnecessary function calls.

internal/events/event_publisher.go
keelerm84: The dropped events metric is a counter, not a gauge, so recording 0 doesn't add any new information and just adds overhead.

internal/events/event_publisher.go
keelerm84: I agree, I should have done this initially. I will do it now.

keelerm84: Updated the events library as well: https://github.com/launchdarkly/go-sdk-events/pull/40

PR #634, opened 2026-04-08
feat: Add request_duration metric for each endpoint

keelerm84:
> [!NOTE]
> **Low Risk**
> Low risk: adds a new OpenCensus duration metric and small middleware timing logic; behavior change is limited to additional metrics emission and explicitly skips unbounded streaming responses.
> 
> **Overview**
> Adds a new `request_duration` metric and view, emitting a distribution of per-request latency tagged by env/platform/user agent/sdk wrapper plus route and method.
> 
> Updates `RequestCount` middleware to time requests and record duration for non-streaming responses, while continuing to count requests and explicitly skipping `text/event-stream` (SSE) endpoints. Tests and shared test exporter utilities are extended to validate distribution metrics and to assert no duration is recorded for streaming routes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b6842d3077f18df39515105f3871834ee8755899. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/metrics/views.go
+++ b/internal/metrics/views.go
@@ -22,6 +22,11 @@ var (
 		Aggregation: view.Count(),
 		TagKeys:     append(publicTags, routeTagKey, methodTagKey),
 	}
+	requestDurationView *view.View = &view.View{ //nolint:gochecknoglobals
+		Measure:     requestDurationMeasure,
+		Aggregation: view.Distribution(10, 25, 50, 100, 250, 500, 1000, 2500, 5000, 10000, 25000, 50000, 100000, 250000, 500000, 1000000),
+		TagKeys:     append(publicTags, routeTagKey, methodTagKey),
+	}
 	privateConnView *view.View = &view.View{ //nolint:gochecknoglobals
 		Measure:     privateConnMeasure,
 		Aggregation: view.Sum(),
@@ -43,7 +48,7 @@ var (
 )
 
 func getPublicViews() []*view.View {
-	return []*view.View{publicConnView, publicNewConnView, requestView}
+	return []*view.View{publicConnView, publicNewConnView, requestView, requestDurationView}
 }
 
 func getPrivateViews() []*view.View {
--- a/internal/middleware/metrics_middleware.go
+++ b/internal/middleware/metrics_middleware.go
@@ -2,6 +2,7 @@ package middleware
 
 import (
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/launchdarkly/ld-relay/v8/internal/metrics"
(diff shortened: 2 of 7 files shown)


internal/metrics/views.go
```diff
@@ -22,6 +22,11 @@ var (
 		Aggregation: view.Count(),
 		TagKeys:     append(publicTags, routeTagKey, methodTagKey),
 	}
+	requestDurationView *view.View = &view.View{ //nolint:gochecknoglobals
+		Measure:     requestDurationMeasure,
+		Aggregation: view.Distribution(10, 25, 50, 100, 250, 500, 1000, 2500, 5000, 10000, 25000, 50000, 100000, 250000, 500000, 1000000),
```
tanderson-ld, line 27: What happens if different versions of relay have different distributions?

internal/middleware/metrics_middleware.go
```diff
@@ -62,9 +64,14 @@ func RequestCount(measure metrics.Measure) mux.MiddlewareFunc {
...
 			// Ignoring internal routing error that would have been ignored anyway
 			route, _ := mux.CurrentRoute(req).GetPathTemplate()
+			start := time.Now()
 			metrics.WithRouteCount(ctx.Env.GetMetricsContext(), userAgent, sdkWrapper, route, req.Method, func() {
 				next.ServeHTTP(w, req)
 			}, measure)
+			// Don't record duration for streaming responses — their lifetime is unbounded
+			if w.Header().Get("X-Accel-Buffering") != "no" {
+				metrics.RecordRequestDuration(ctx.Env.GetMetricsContext(), userAgent, sdkWrapper, route, req.Method, time.Since(start), measure)
+			}
```
tanderson-ld, line 75: Is this a long term stable way to make this distinction? Does the application have the logical context to provide a parameter that says this is long lived, or perhaps something on the request reference?

A quick google search on that header seems to indicate this relates to NGINX.  Seems a little off to make this decision based on a header respected by a certain popular proxy.

keelerm84: The relay always adds that header for streaming requests, which is why it's okay for us to check against it.

internal/metrics/views.go
keelerm84: Different distributions will cause an individual metric to potentially be classified differently.

As is, a 23 microsecond response time gets bucketed into the second bucket. If we defined this differently, perhaps it would go into bucket 2 or 3.

This could lead to your P50/P90/etc showing a little differently. But as long as the thresholds aren't insanely different, it shouldn't be a big deal.

In v9 with OTEL, we move to dynamic, exponential distributions which is the new recommendation.

internal/middleware/metrics_middleware.go
kinyoklion: I am fine with this, but I do think "Content-Type: text/event-stream" is better. SSE will always have it, and it directly says that it is a stream. So even if you used it for non-sse we would still want to exclude it. (This is the header that has been used to fix the "wait for complete" logic in the various network debug tools.)

kinyoklion (approved):

PR #701, opened 2026-06-17
feat(credential): remove RotateWithGrace mobile-key panic

aaron-zeisler:
## Summary
Removes the panic guard at the `MobileKey` branch of `Rotator.RotateWithGrace` and adds a `deprecatedMobileKeys` map to store the outgoing key's expiry when a grace period is provided.

[Jira: SDK-2536](https://launchdarkly.atlassian.net/browse/SDK-2536)

## Background
The panic existed because the Rotator had no data-model slot for an expiring mobile key — failing loud was preferable to silently storing junk. T1.a (next) will add the full accepted-set data structures; this PR removes the now-incorrect guard ahead of that, and adds the minimal storage needed to hold the expiry time. `StepTime` does not yet act on `deprecatedMobileKeys` — that is deferred to T1.c (generalize cleanup ticker).

Behavior for all existing callers (grace == nil) is unchanged.

---

> [!NOTE]
> **Medium Risk**
> Touches credential rotation for mobile keys (auth-related), but existing nil-grace call paths are unchanged and grace-period keys are not yet enforced by `StepTime` or exposed via `DeprecatedCredentials`.
> 
> **Overview**
> **Mobile key rotation** no longer panics when `RotateWithGrace` is called with a non-nil grace period. The outgoing key can be recorded in a new `deprecatedMobileKeys` map (key → expiry), mirroring the SDK-key deprecation model at a minimal level.
> 
> `updateMobileKey` now takes `grace *GracePeriod`: **nil** or **already-expired** grace still revokes the previous key immediately (same as before for existing callers). A valid grace period only stores the old key’s expiry; **`StepTime` does not expire deprecated mobile keys yet** (planned in T1.c). Re-promoting a key clears it from `deprecatedMobileKeys`.
> 
> Tests cover the non-panic grace path, immediate revoke cases, and re-promotion out of the deprecated set.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f9f8e4999877bd9364f810b114282b3cfad116d5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/credential/rotator.go
+++ b/internal/credential/rotator.go
@@ -12,8 +12,13 @@ import (
 type Rotator struct {
 	loggers ldlog.Loggers
 
-	// There is only one mobile key active at a given time; it does not support a deprecation period.
+	// There is only one mobile key active at a given time.
 	primaryMobileKey config.MobileKey
+
+	// deprecatedMobileKeys stores mobile keys being phased out with a grace period, keyed
+	// by credential value with the associated expiry time. StepTime does not yet act on
+	// these entries — that is deferred to T1.c (generalize cleanup ticker).
+	deprecatedMobileKeys map[config.MobileKey]time.Time
 
 	// There is only one environment ID active at a given time, and it won't actually be rotated. The mechanism is
 	// here to allow setting it in a deferred manner.
@@ -42,8 +47,9 @@ type InitialCredentials struct {
 // contains no credentials and can optionally be initialized via Initialize.
 func NewRotator(loggers ldlog.Loggers) *Rotator {
 	r := &Rotator{
-		loggers:           loggers,
-		deprecatedSdkKeys: make(map[config.SDKKey]time.Time),
+		loggers:              loggers,
+		deprecatedSdkKeys:    make(map[config.SDKKey]time.Time),
+		deprecatedMobileKeys: make(map[config.MobileKey]time.Time),
 	}
 	return r
 }
@@ -158,17 +164,19 @@ func NewGracePeriod(key config.SDKKey, expiry time.Time, now time.Time) *GracePe
 
 // RotateWithGrace sets a new primary credential while deprecating a previous credential. The grace
 // parameter may be nil to immediately revoke the previous credential.
-// It is invalid to specify a grace period when the credential being rotate is a mobile key or
-// environment ID.
+// It is invalid to specify a grace period when the credential being rotated is an environment ID.
+// For mobile keys, a non-nil grace period stores the expiry for the outgoing key; the cleanup
+// ticker (T1.c) is responsible for acting on it.
 func (r *Rotator) RotateWithGrace(primary SDKCredential, grace *GracePeriod) {
 	switch primary := primary.(type) {
 	case config.SDKKey:
 		r.updateSDKKey(primary, grace)
 	case config.MobileKey:
 		if grace != nil {
-			panic("programmer error: mobile keys do not support deprecation")
+			r.updateMobileKeyWithGrace(primary, grace)
+		} else {
+			r.updateMobileKey(primary)
 		}
-		r.updateMobileKey(primary)
 	case config.EnvironmentID:
 		if grace != nil {
 			panic("programmer error: environment IDs do not support deprecation")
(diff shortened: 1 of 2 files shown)


internal/credential/rotator.go
```diff
@@ -211,6 +219,29 @@ func (r *Rotator) updateMobileKey(mobileKey config.MobileKey) {
 	}
 }
 
+func (r *Rotator) updateMobileKeyWithGrace(mobileKey config.MobileKey, grace *GracePeriod) {
+	if mobileKey == r.MobileKey() {
```
keelerm84: I don't think this check should exist outside of the lock check.

internal/credential/rotator.go
```diff
@@ -158,17 +164,19 @@ func NewGracePeriod(key config.SDKKey, expiry time.Time, now time.Time) *GracePe
...
 func (r *Rotator) RotateWithGrace(primary SDKCredential, grace *GracePeriod) {
 	switch primary := primary.(type) {
 	case config.SDKKey:
 		r.updateSDKKey(primary, grace)
 	case config.MobileKey:
 		if grace != nil {
-			panic("programmer error: mobile keys do not support deprecation")
+			r.updateMobileKeyWithGrace(primary, grace)
+		} else {
+			r.updateMobileKey(primary)
```
aaron-zeisler: I think it's better to follow the same pattern as `updateSDKKey()`, where there's only a single function that takes both the key and the grace period, and handling a nil grace period is dealt with inside the function. I think that's a better situation than having two different functions: one that takes `grace` and one that does not.

internal/credential/rotator.go
```diff
@@ -211,6 +219,29 @@ func (r *Rotator) updateMobileKey(mobileKey config.MobileKey) {
...
+	if grace.Expired() {
+		r.loggers.Infof("Deprecated mobile key %s already expired at %v; revoking immediately", previous.Masked(), grace.expiry)
+		r.expirations = append(r.expirations, previous)
+		return
+	}
+	r.deprecatedMobileKeys[previous] = grace.expiry
+	r.loggers.Infof("Mobile key %s was marked for deprecation with an expiry at %v, new primary mobile key is %s",
+		previous.Masked(), grace.expiry, mobileKey.Masked())
+}
+
```
aaron-zeisler, line 231: This code could use comments: at least one godoc-style comment decorating the function.

internal/credential/rotator_test.go
```diff
@@ -231,3 +231,53 @@ func TestSDKKeyExpiredInThePastIsNotAdded(t *testing.T) {
...
+
+		rotator.RotateWithGrace(mob2, NewGracePeriod(config.SDKKey(""), expiry, now))
+
+		assert.Equal(t, mob2, rotator.MobileKey())
+
+		additions, expirations := rotator.StepTime(now)
+		assert.ElementsMatch(t, []SDKCredential{mob2}, additions)
+		assert.ElementsMatch(t, []SDKCredential{mob1}, expirations)
+	})
+}
```
aaron-zeisler, line 327: After we combine the two functions into one as described above, we'll need to extend this test (and possibly move code from an existing test, or consolidate tests).

internal/credential/rotator.go
```diff
@@ -156,19 +162,16 @@ func NewGracePeriod(key config.SDKKey, expiry time.Time, now time.Time) *GracePe
...
 }
 
-// RotateWithGrace sets a new primary credential while deprecating a previous credential. The grace
-// parameter may be nil to immediately revoke the previous credential.
-// It is invalid to specify a grace period when the credential being rotate is a mobile key or
-// environment ID.
+// RotateWithGrace sets a new primary credential while deprecating the previous one. When grace is nil
+// the outgoing credential is immediately revoked. It is invalid to specify a grace period for an
+// environment ID. For mobile keys, a non-nil grace period stores the expiry for the outgoing key;
+// the cleanup ticker (T1.c) is responsible for acting on it.
```
keelerm84:
```suggestion
// the cleanup ticker is responsible for acting on it.
```

keelerm84 (approved):

PR #702, opened 2026-06-17
feat(envfactory): extend EnvironmentRep and EnvironmentParams for concurrent keys

aaron-zeisler:
## Summary

Implements [SDK-2545](https://launchdarkly.atlassian.net/browse/SDK-2545) — the "data first" PR for the concurrent-keys wire-format side. No consumers yet (those land in T3.b/c).

- Adds `ConcurrentKeyRep` wire type for entries in the `sdkKeys`/`mobileKeys` arrays
- Adds `SDKKeys []ConcurrentKeyRep` and `MobileKeys []ConcurrentKeyRep` to `EnvironmentRep`
- Adds `AcceptedSDKKey`, `AcceptedMobileKey`, and corresponding slice fields to `EnvironmentParams`
- `EnvironmentRep.ToParams()` populates the new accepted-set fields from the arrays when present and non-empty; absent or empty arrays are synthesized from the singular `sdkKey`/`mobKey` fields so downstream consumers always receive a non-empty accepted set with at least the anchor
- Adds canonical TERMINOLOGY comment to `EnvironmentRep`

## Background

Part of the Phase 1 concurrent SDK keys feature ([SDK-2453](https://launchdarkly.atlassian.net/browse/SDK-2453)). Both RAC and the offline archive carry `EnvironmentRep`; one parsing change covers both sources.

Design reference: `docs/concurrent-keys/phase1-design.md` §5 (Wire format) and §10 (Backwards compatibility).

## Pre-work verifications (T3.a acceptance criteria)

- **`DisallowUnknownFields` absent**: confirmed by grep — not used anywhere in the env-parse path (`internal/envfactory/`, `internal/filedata/`, `relay/`). Documented in the new `EnvironmentRep` comment.
- **`archive_reader.go` uses `EnvironmentRep` directly**: confirmed — `archiveEnvironmentRep.Env` is `envfactory.EnvironmentRep` and calls `rep.Env.ToParams()`. No scope expansion needed.

## Changes

- `internal/envfactory/env_rep.go` — new `ConcurrentKeyRep` type; `SDKKeys`/`MobileKeys` fields on `EnvironmentRep`; canonical TERMINOLOGY comment; `ToParams()` with new-format (non-empty array) and old-format/empty synthesis branches
- `internal/envfactory/env_params.go` — new `AcceptedSDKKey`, `AcceptedMobileKey` types; `AcceptedSDKKeys`/`AcceptedMobileKeys` fields on `EnvironmentParams`
- `internal/envfactory/env_rep_test.go` — tests covering new-format parsing and old-format normalization

## Test coverage

| Test | What it verifies |
|---|---|
| `TestEnvironmentRepToParams` | `ToParams()` output matches expected params for single key and expiring key cases |
| `TestEnvironmentRepJSONFormat` | Legacy wire format (singular `sdkKey`) deserializes correctly |
| `TestEnvironmentRepNewFormatWithArrays` | Parse realistic RAC `put` payload with `sdkKeys`/`mobileKeys`; struct fields + `ToParams()` output |
| `TestEnvironmentRepOldFormatNoArrays` | Old-format payload (no arrays) → `ToParams()` synthesizes non-nil `AcceptedSDKKeys`/`AcceptedMobileKeys` from singular fields |

All existing tests pass unchanged.

[SDK-2545]: https://launchdarkly.atlassian.net/browse/SDK-2545?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/envfactory/env_params.go
+++ b/internal/envfactory/env_params.go
@@ -49,7 +49,7 @@ type EnvironmentParams struct {
 // AcceptedSDKKey is one entry in the accepted SDK key set for an environment.
 // Expiry is zero if the key is permanent.
 type AcceptedSDKKey struct {
-	Identifier string
+	Key string
 	Value      config.SDKKey
 	Expiry     time.Time
 }
@@ -57,7 +57,7 @@ type AcceptedSDKKey struct {
 // AcceptedMobileKey is one entry in the accepted mobile key set for an environment.
 // Expiry is zero if the key is permanent.
 type AcceptedMobileKey struct {
-	Identifier string
+	Key string
 	Value      config.MobileKey
 	Expiry     time.Time
 }
--- a/internal/envfactory/env_params.go
+++ b/internal/envfactory/env_params.go
@@ -49,17 +49,17 @@ type EnvironmentParams struct {
 // AcceptedSDKKey is one entry in the accepted SDK key set for an environment.
 // Expiry is zero if the key is permanent.
 type AcceptedSDKKey struct {
-	Key string
-	Value      config.SDKKey
-	Expiry     time.Time
+	Key    string
+	Value  config.SDKKey
+	Expiry time.Time
 }
 
 // AcceptedMobileKey is one entry in the accepted mobile key set for an environment.
 // Expiry is zero if the key is permanent.
 type AcceptedMobileKey struct {
-	Key string
-	Value      config.MobileKey
-	Expiry     time.Time
+	Key    string
+	Value  config.MobileKey
+	Expiry time.Time
 }
 
 type ExpiringSDKKey struct {
--- a/internal/envfactory/env_params.go
+++ b/internal/envfactory/env_params.go
@@ -31,13 +31,14 @@ type EnvironmentParams struct {
 	ExpiringSDKKey ExpiringSDKKey
 
 	// AcceptedSDKKeys is the full accepted set of SDK keys for this environment, including the
-	// anchor. Always non-nil after ToParams(): new-format payloads populate from the sdkKeys
-	// array; old-format payloads (no sdkKeys field) are synthesized from the singular sdkKey field.
+	// anchor. Always non-nil after ToParams(): non-empty sdkKeys arrays populate directly; absent
+	// or empty sdkKeys are synthesized from the singular sdkKey field so there is always at least
+	// the anchor entry.
 	AcceptedSDKKeys []AcceptedSDKKey
 
 	// AcceptedMobileKeys is the full accepted set of mobile keys for this environment. Always
-	// non-nil after ToParams(): new-format payloads populate from the mobileKeys array; old-format
-	// payloads are synthesized from the singular mobKey field.
+	// non-nil after ToParams(): non-empty mobileKeys arrays populate directly; absent or empty
+	// mobileKeys are synthesized from the singular mobKey field.
 	AcceptedMobileKeys []AcceptedMobileKey
 
 	// TTL is the cache TTL for PHP clients.
--- a/internal/envfactory/env_params.go
+++ b/internal/envfactory/env_params.go
@@ -31,12 +31,13 @@ type EnvironmentParams struct {
 	ExpiringSDKKey ExpiringSDKKey
 
 	// AcceptedSDKKeys is the full accepted set of SDK keys for this environment, including the
-	// anchor. Populated only when the source (RAC or offline archive) emits the sdkKeys array.
-	// Nil for old-format payloads that carry only the singular sdkKey field.
+	// anchor. Always non-nil after ToParams(): new-format payloads populate from the sdkKeys
+	// array; old-format payloads (no sdkKeys field) are synthesized from the singular sdkKey field.
 	AcceptedSDKKeys []AcceptedSDKKey
 
-	// AcceptedMobileKeys is the full accepted set of mobile keys for this environment. Populated
-	// only when the source emits the mobileKeys array; nil for old-format payloads.
+	// AcceptedMobileKeys is the full accepted set of mobile keys for this environment. Always
+	// non-nil after ToParams(): new-format payloads populate from the mobileKeys array; old-format
+	// payloads are synthesized from the singular mobKey field.
 	AcceptedMobileKeys []AcceptedMobileKey
 
 	// TTL is the cache TTL for PHP clients.
--- a/internal/envfactory/env_params.go
+++ b/internal/envfactory/env_params.go
@@ -30,11 +30,36 @@ type EnvironmentParams struct {
 	// the canonical one).
 	ExpiringSDKKey ExpiringSDKKey
 
+	// AcceptedSDKKeys is the full accepted set of SDK keys for this environment, including the
+	// anchor. Populated only when the source (RAC or offline archive) emits the sdkKeys array.
+	// Nil for old-format payloads that carry only the singular sdkKey field.
+	AcceptedSDKKeys []AcceptedSDKKey
+
+	// AcceptedMobileKeys is the full accepted set of mobile keys for this environment. Populated
+	// only when the source emits the mobileKeys array; nil for old-format payloads.
+	AcceptedMobileKeys []AcceptedMobileKey
+
 	// TTL is the cache TTL for PHP clients.
 	TTL time.Duration
 
 	// SecureMode is true if secure mode is required for this environment.
 	SecureMode bool
+}
+
+// AcceptedSDKKey is one entry in the accepted SDK key set for an environment.
+// Expiry is zero if the key is permanent.
+type AcceptedSDKKey struct {
+	Identifier string
+	Value      config.SDKKey
+	Expiry     time.Time
+}
+
+// AcceptedMobileKey is one entry in the accepted mobile key set for an environment.
+// Expiry is zero if the key is permanent.
+type AcceptedMobileKey struct {
+	Identifier string
+	Value      config.MobileKey
+	Expiry     time.Time
 }
 
 type ExpiringSDKKey struct {
--- a/internal/envfactory/env_rep.go
+++ b/internal/envfactory/env_rep.go
@@ -35,19 +35,21 @@ import (
 //
 // TERMINOLOGY:
 //
-//	The wire "key" field is the human-readable IDENTIFIER (e.g. "default-sdk"),
-//	non-secret. The wire "value" field is the actual CREDENTIAL string (e.g.
-//	"sdk-xxxx-..."), which is the secret. Note that relay's own types
-//	(SDKKey, MobileKey, SDKCredential) refer to what the wire calls "value" —
-//	they are misnamed by today's standards but stable, so do not rename them.
+//	The wire "key" field is the human-readable identifier (e.g. "default-sdk"),
+//	non-secret — stored as AcceptedSDKKey.Key / AcceptedMobileKey.Key internally.
+//	The wire "value" field is the actual credential string (e.g. "sdk-xxxx-..."),
+//	which is the secret — stored as AcceptedSDKKey.Value / AcceptedMobileKey.Value.
+//	Note that relay's own types (SDKKey, MobileKey, SDKCredential) refer to what
+//	the wire calls "value" — they are misnamed by today's standards but stable,
+//	so do not rename them.
 //
 // Anchor selection: anchor = the sdkKeys entry whose `value` matches sdkKey.value.
-// No isDefault flag — value match is the signal. See phase1-design.md §4.2.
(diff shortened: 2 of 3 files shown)


internal/envfactory/env_params.go
```diff
@@ -30,13 +30,38 @@ type EnvironmentParams struct {
...
 	TTL time.Duration
 
 	// SecureMode is true if secure mode is required for this environment.
 	SecureMode bool
 }
 
+// AcceptedSDKKey is one entry in the accepted SDK key set for an environment.
+// Expiry is zero if the key is permanent.
+type AcceptedSDKKey struct {
+	Identifier string
```
keelerm84: The RAC stream refers to this as `key`.

internal/envfactory/env_params.go
```diff
@@ -30,13 +30,38 @@ type EnvironmentParams struct {
...
+	Value      config.SDKKey
+	Expiry     time.Time
+}
+
+// AcceptedMobileKey is one entry in the accepted mobile key set for an environment.
+// Expiry is zero if the key is permanent.
+type AcceptedMobileKey struct {
+	Identifier string
```
keelerm84: Same here re: `key`.

internal/envfactory/env_rep.go
```diff
@@ -15,6 +15,39 @@ import (
...
+// TERMINOLOGY:
+//
+//	The wire "key" field is the human-readable IDENTIFIER (e.g. "default-sdk"),
+//	non-secret. The wire "value" field is the actual CREDENTIAL string (e.g.
+//	"sdk-xxxx-..."), which is the secret. Note that relay's own types
+//	(SDKKey, MobileKey, SDKCredential) refer to what the wire calls "value" —
+//	they are misnamed by today's standards but stable, so do not rename them.
+//
+// Anchor selection: anchor = the sdkKeys entry whose `value` matches sdkKey.value.
+// No isDefault flag — value match is the signal. See phase1-design.md §4.2.
```
keelerm84:
```suggestion
// No isDefault flag — value match is the signal.
```

internal/envfactory/env_rep.go
```diff
@@ -15,6 +15,39 @@ import (
...
+//
+// Backwards compatibility: Go's default JSON decoder ignores unknown fields, so old
+// relays receiving payloads with sdkKeys/mobileKeys simply ignore them and continue
+// using sdkKey/mobKey. DisallowUnknownFields is intentionally not used anywhere in
+// this parse path — verified as T3.a pre-work. See phase1-design.md §10.
```
keelerm84:
```suggestion
// this parse path.
```

internal/envfactory/env_rep_test.go
```diff
@@ -105,3 +105,119 @@ func TestEnvironmentRepJSONFormat(t *testing.T) {
...
+		Identifier: "service-a",
+		Value:      config.SDKKey("sdk-service-a"),
+		Expiry:     time.UnixMilli(expiryMs),
+	}, params.AcceptedSDKKeys[1])
+
+	require.Len(t, params.AcceptedMobileKeys, 1)
+	assert.Equal(t, AcceptedMobileKey{Identifier: "mob-key-1", Value: config.MobileKey("mob-f41c")}, params.AcceptedMobileKeys[0])
+}
+
+// TestEnvironmentRepOldFormatNoArrays verifies that an old-format payload (singular sdkKey/mobKey
```
keelerm84, line 166: This is an interesting test case, and worth considering more.

Since the RAC format has already been modified publicly, we don't really need to worry about the old format there.

However, a customer might have an older offline mode file, so we should maintain some type of backwards compatibility there. However, I don't want our implementation to have to consider both formats.

Is there some post JSON decoding work we can do to ensure we have a consistent internal model to operate on?

aaron-zeisler: Agreed — normalization in `ToParams()` is the right move. Here's what I'm thinking:

If `SDKKeys` is nil (old-format payload), synthesize `AcceptedSDKKeys` from the singular fields rather than leaving it nil:
- Add an `AcceptedSDKKey` for `sdkKey.value` (the anchor)
- If `sdkKey.expiring` is present, fold that in too with its expiry timestamp

Same pattern for mobile keys from `mobKey`.

Result: T3.b/c always gets a non-nil `AcceptedSDKKeys` with at least the anchor entry — no nil-check, no format-detection logic in consumers.

One detail: synthesized entries from old-format payloads would have `Key: ""` since the old wire format has no identifier field. That's honest (there's nothing to invent), and safe because reconcile logic matches by `Value`, not `Key` — identifiers are display-only (used in T4 status output).

With that in place, `TestEnvironmentRepOldFormatNoArrays` would flip from asserting `AcceptedSDKKeys == nil` to asserting a synthesized single-entry slice containing the anchor. I'll implement this now if you're good with the approach.

aaron-zeisler: @keelerm84 I went ahead and made the suggested fix, and it's here in the 3rd commit. Let me know if it looks good or if we should try a different approach.

internal/envfactory/env_rep.go
```diff
@@ -94,6 +142,53 @@ func (r EnvironmentRep) ToParams() EnvironmentParams {
...
+				Value: config.SDKKey(k.Value),
+			}
+			if k.Expiry != nil {
+				entry.Expiry = time.UnixMilli(*k.Expiry)
+			}
+			params.AcceptedSDKKeys = append(params.AcceptedSDKKeys, entry)
+		}
+	} else {
+		// Old-format payload: no sdkKeys array present. Synthesize AcceptedSDKKeys from the
+		// singular sdkKey fields so T3.b/c consumers always receive a consistent non-nil model
```
keelerm84:
```suggestion
		// singular sdkKey fields so consumers always receive a consistent non-nil model
```

internal/envfactory/env_rep.go
```diff
@@ -94,6 +142,53 @@ func (r EnvironmentRep) ToParams() EnvironmentParams {
...
+		// regardless of wire format version. Key (identifier) is empty — the old format had none.
```
keelerm84, line 161: Do we think this is going to matter?

internal/envfactory/env_rep_test.go
```diff
@@ -35,10 +35,12 @@ func TestEnvironmentRepToParams(t *testing.T) {
...
-		SDKKey:     env1.SDKKey.Value,
-		MobileKey:  env1.MobKey,
-		TTL:        2 * time.Minute,
-		SecureMode: true,
+		SDKKey:             env1.SDKKey.Value,
+		MobileKey:          env1.MobKey,
+		TTL:                2 * time.Minute,
+		SecureMode:         true,
+		AcceptedSDKKeys:    []AcceptedSDKKey{{Value: config.SDKKey("sdkkey1")}},
+		AcceptedMobileKeys: []AcceptedMobileKey{{Value: config.MobileKey("mobkey1")}},
```
keelerm84: Here and in other places I'm sure, this should be more like the following to ensure they stay consistent.


```suggestion
		AcceptedSDKKeys:    []AcceptedSDKKey{{Value: env1.SDKKey.Value}},
		AcceptedMobileKeys: []AcceptedMobileKey{{Value: env1.MobKey}},
```

internal/envfactory/env_rep_test.go
```diff
@@ -105,3 +112,150 @@ func TestEnvironmentRepJSONFormat(t *testing.T) {
...
+	}, params.AcceptedSDKKeys[1])
+
+	require.Len(t, params.AcceptedMobileKeys, 1)
+	assert.Equal(t, AcceptedMobileKey{Key: "mob-key-1", Value: config.MobileKey("mob-f41c")}, params.AcceptedMobileKeys[0])
+}
+
+// TestEnvironmentRepOldFormatNoArrays verifies that an old-format payload (singular sdkKey/mobKey
+// only, no sdkKeys/mobileKeys arrays) is normalized by ToParams() into a consistent accepted set.
+// The wire rep's SDKKeys/MobileKeys remain nil, but params.AcceptedSDKKeys/AcceptedMobileKeys are
+// synthesized from the singular fields so T3.b/c consumers never need to handle two code paths.
```
keelerm84:
```suggestion
// synthesized from the singular fields so consumers never need to handle two code paths.
```

internal/envfactory/env_rep_test.go
```diff
@@ -105,3 +112,150 @@ func TestEnvironmentRepJSONFormat(t *testing.T) {
...
+	assert.Equal(t, AcceptedSDKKey{Value: config.SDKKey("sdk-key1")}, params.AcceptedSDKKeys[0])
+	require.Len(t, params.AcceptedMobileKeys, 1)
+	assert.Equal(t, AcceptedMobileKey{Value: config.MobileKey("mob-default")}, params.AcceptedMobileKeys[0])
+}
+
+// TestEnvironmentRepEmptyArrays verifies that a new-format payload carrying explicitly empty
+// sdkKeys/mobileKeys arrays (present but []) produces non-nil empty AcceptedSDKKeys/
+// AcceptedMobileKeys — distinct from the old-format synthesis path which always produces at
+// least one entry (the anchor).
+func TestEnvironmentRepEmptyArrays(t *testing.T) {
```
keelerm84: This should be impossible, given the way upstream constructs this. But I'm not entirely sure we want to ignore potentially useful information.

Maybe we should consider those fields being nil or empty as the same condition in `ToParams`. What do you thinK?

internal/envfactory/env_rep_test.go
```diff
@@ -105,3 +112,150 @@ func TestEnvironmentRepJSONFormat(t *testing.T) {
...
+	assert.NotNil(t, params.AcceptedSDKKeys, "non-nil empty sdkKeys must produce non-nil AcceptedSDKKeys")
+	assert.Len(t, params.AcceptedSDKKeys, 0)
+	assert.NotNil(t, params.AcceptedMobileKeys, "non-nil empty mobileKeys must produce non-nil AcceptedMobileKeys")
+	assert.Len(t, params.AcceptedMobileKeys, 0)
+}
+
+// TestEnvironmentRepBackCompatOldRelay verifies that a new-format payload (containing sdkKeys/
+// mobileKeys) is parsed cleanly by a relay that does not know about those fields. Go's JSON
+// decoder ignores unknown fields by default; this test makes that guarantee explicit and durable.
+// DisallowUnknownFields is intentionally not used in this parse path (see phase1-design.md §10).
```
keelerm84:
```suggestion
// DisallowUnknownFields is intentionally not used in this parse path.
```

Might want to add something to your prompt about not including this sort of thing.

internal/envfactory/env_rep_test.go
```diff
@@ -105,3 +112,150 @@ func TestEnvironmentRepJSONFormat(t *testing.T) {
...
+	assert.Len(t, rep.MobileKeys, 0)
+
+	params := rep.ToParams()
+	assert.NotNil(t, params.AcceptedSDKKeys, "non-nil empty sdkKeys must produce non-nil AcceptedSDKKeys")
+	assert.Len(t, params.AcceptedSDKKeys, 0)
+	assert.NotNil(t, params.AcceptedMobileKeys, "non-nil empty mobileKeys must produce non-nil AcceptedMobileKeys")
+	assert.Len(t, params.AcceptedMobileKeys, 0)
+}
+
+// TestEnvironmentRepBackCompatOldRelay verifies that a new-format payload (containing sdkKeys/
```
keelerm84: This isn't a test that should be kept. Makes no sense for new versions to test that old versions work.

internal/envfactory/env_rep.go
aaron-zeisler: @keelerm84 with this comment, are you referring specifically to "Key (identifier) is empty — the old format had none"? Do you think we should populate the key anyways?

Or are you wondering about the entire "old-format payload" condition?

Please elaborate 🙂

aaron-zeisler: OK, I think the risk is very low.

For RAC, I think this would only happen if the backend sent the old-style payload. Since it's already serving the new-style payload, I think this would only happen if there was a rollback for the backend services.

Ah, this is a probable situation for "offline mode", though.

OK, consulting w/ Claude, it seems that this won't cause auth to break. The risk would be if the key was exposed in some way, like in a log, or if the list of accepted keys is returned from a function. I don't think we need to do those actions - I can try to prevent those situations from being necessary throughout this project, or deal with the issue if/when it actually comes up.

keelerm84 (approved):

PR #703, opened 2026-06-17
test(events): add credential routing tests for EventDispatcher

aaron-zeisler:
## Summary

Adds three tests covering the credential routing invariants of `EventDispatcher` that are not exercised by existing tests.

[Jira: SDK-2549](https://launchdarkly.atlassian.net/browse/SDK-2549)

## Background

The existing `TestVerbatimEventHandlers`, `TestSummarizeEvents`, and `TestDiagnosticEventForwarding` tests confirm schema correctness but all configure the same credential on both the incoming request and the dispatcher. They don't exercise the case where an incoming SDK request carries a credential that differs from the dispatcher's stored `authKey` (the anchor). The new tests specifically set a non-anchor `Authorization` on the incoming request so the routing behavior is directly asserted.

## Changes

Three tests in `internal/events/event_payload_regression_test.go`:

- **`TestAnalyticsUpstreamUsesAnchorCredential`** — incoming request carries a non-anchor SDK key; asserts the upstream request uses the dispatcher's anchor credential, not the incoming one.
- **`TestDiagnosticUpstreamProxiesIncomingCredential`** — asserts the upstream diagnostic request preserves the incoming `Authorization` header verbatim and does not use the anchor.
- **`TestCredentialRoutingAfterReplaceCredential`** — calls `ReplaceCredential` and asserts that analytics switches to the new anchor immediately while diagnostic forwarding continues to proxy the original incoming credential.

---

> [!NOTE]
> **Low Risk**
> Test-only change with no production code modifications; it documents and guards existing credential routing behavior.
> 
> **Overview**
> Adds **`internal/events/event_payload_regression_test.go`** with regression tests for `EventDispatcher` **Authorization** routing when the incoming SDK request’s credential differs from the dispatcher’s stored anchor (`authKey`).
> 
> **Analytics** upstream requests must always use the anchor, not the client’s header (`TestAnalyticsUpstreamUsesAnchorCredential`). **Diagnostic** upstream requests must forward the incoming `Authorization` verbatim and never substitute the anchor (`TestDiagnosticUpstreamProxiesIncomingCredential`). After **`ReplaceCredential`**, analytics must use the new anchor immediately while diagnostic behavior stays unchanged (`TestCredentialRoutingAfterReplaceCredential`).
> 
> Tests reuse the existing `eventRelayTest` harness and assert upstream headers via a captured HTTP channel.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b1a12d44e36cacfbd35ec1ec03b218365a790d82. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


docs/concurrent-keys/baseline-refresh.md
```diff
@@ -0,0 +1,60 @@
+# How to Refresh the Event Payload Regression Baseline
```
keelerm84: This doesn't belong in the docs folder. The docs folder is for public consumption. If you mean this for devs, it should be in a README in that folder. Or if it's for agentic use, we should put it in a CLAUDE.md file within that folder.

I'm skeptical whether or not this is actually needed though. At a bare minimum, this should be cut WAY done. We shouldn't reference temporary artifacts like "phase 1" in anything that will be permanent in the repo.

internal/events/event_payload_regression_test.go
```diff
@@ -0,0 +1,309 @@
+// Event payload regression tests for concurrent-keys Phase 1.
```
keelerm84: Same comment here about the phase 1 mention.

internal/events/event_payload_regression_test.go
```diff
@@ -0,0 +1,309 @@
...
+	require.NoError(t, json.Unmarshal(fixture, &fixtureVal), "fixture must be valid JSON")
+	require.NoError(t, json.Unmarshal(actual, &actualVal), "upstream body must be valid JSON")
+
+	// For event-batch arrays, compare element-by-element so every event kind is verified.
+	fixtureArr, fixtureIsArr := fixtureVal.([]any)
+	actualArr, actualIsArr := actualVal.([]any)
+	if fixtureIsArr && actualIsArr {
+		assert.Equal(t, len(fixtureArr), len(actualArr),
+			"event array length must match baseline")
+		n := min(len(fixtureArr), len(actualArr))
```
keelerm84: Shouldn't need the min since we assert they are the same length above. Could just pick one.

internal/events/event_payload_regression_test.go
```diff
@@ -0,0 +1,309 @@
...
+		assert.Equal(t, 202, w.Result().StatusCode)
+
+		p.dispatcher.flush()
+		r := expectUpstreamRequest(t, p.requestsCh)
+
+		// Invariant 1: analytics upstream must carry the anchor credential.
+		assert.Equal(t, string(st.EnvMain.Config.SDKKey), r.Request.Header.Get("Authorization"),
+			"analytics upstream must use anchor credential, not the incoming request credential")
+
+		// Invariant 4: body schema must match v8 baseline.
```
keelerm84: I don't think these schema checks are really that important. They add a lot of noise to what would otherwise be an important test -- that the SDK key used by the connecting SDK doesn't always align with the key we are using to send the events.

There should be plenty of other tests already covering that the schema is behaving the way we need.

docs/concurrent-keys/baseline-refresh.md
aaron-zeisler: I think the right place for this file is in docs/concurrent-keys/ - this is where I'm storing files that are specific to the concurrent keys project, and files that are meant for agents. I'm planning to remove that entire directory before we merge the code back into v8.

aaron-zeisler:
General question: Are these tests useful after the Concurrent Keys project ships? At the end of the project, should these test be removed? Or moved to another permanent place (possibly re-written) where they'll be useful for the foreseeable future.

internal/events/event_payload_regression_test.go
aaron-zeisler: I'm going to remove the schema checks - like you mentioned, I see other tests that are validating the same thing. I'll rewrite these tests so they're useful after tge concurrent keys project (and I'll remove any mention of the project from the tests).

docs/concurrent-keys/baseline-refresh.md
aaron-zeisler: I dropped this entire file

keelerm84 (approved):

PR #705, opened 2026-06-17
test(sharedtest): add Phase 1 integration test utilities

aaron-zeisler:
## Summary
Adds reusable Phase 1 concurrent-key test helpers to `internal/sharedtest`, plus a reference end-to-end test demonstrating them. Rather than introducing a separate test package, the helpers extend the existing `sharedtest` framework so they're importable across packages without a parallel framework.

[Jira: SDK-2548](https://launchdarkly.atlassian.net/browse/SDK-2548)

## Background
Phase 1 scenario tests (T1.c, T2.c, T3.c, T4) need to exercise Relay end-to-end under both auto-config (RAC) and offline-mode (filedata) paths and assert on the events a downstream SDK receives. The relevant patterns already existed as one-off helpers inside `package relay` test files (`autoConfTest`, the archive builders) and in `sharedtest` (`WithStreamRequest`), but `_test.go` helpers in `package relay` aren't importable from other packages' tests. This PR extracts the genuinely-new pieces into `sharedtest` so the distributed sub-task tests can share them.

## Changes
- **`internal/sharedtest/configsource`** (new subpackage) — `RACMock` + `MakeAutoConfig{Put,Patch,Delete}Event` (RAC SSE mock) and `ArchiveFixtureBuilder` (offline `.tar.gz` builder). These reference `envfactory`, which transitively imports `relayenv`/`streams`, so they live in a subpackage — mirroring the existing `testclient`/`testenv` pattern — to keep top-level `sharedtest` importable by `relayenv` and `streams` without an import cycle.
- **`internal/sharedtest/stream_request.go`** — adds `AwaitEventOfType`, a small matcher over the event channel that `WithStreamRequest` already provides. No new SSE consumer was introduced; the downstream-SDK side reuses `WithStreamRequest` (in-process, deterministic teardown).
- **`relay/concurrent_keys_harness_ref_test.go`** — reference end-to-end test with two sub-tests (offline archive path and RAC path), using the helpers above.

## How upcoming tasks will use these
Each Phase 1 sub-task brings its own scenario tests; this PR just gives them the shared building blocks. In plain terms:

- **T3.c — wire RAC + offline handlers** (the heaviest user). This task makes Relay accept multiple keys from RAC and from offline archives. Its tests will use **`RACMock`** to play a sequence of config events at Relay — e.g. an initial `put`, then a `patch` that adds a second SDK key, then a `delete` — and use **`ArchiveFixtureBuilder`** to build the equivalent offline archive for the filedata path. After each change they'll connect with **`WithStreamRequest`** and use **`AwaitEventOfType`** to confirm the downstream SDK stream reflects the reconciled set of keys. This covers the reconcile matrix: add a key, set a key to expire, remove, rename, un-expire, mixed changes, and partial-failure cases.

- **T4 — status endpoint arrays.** This task exposes the multiple keys on Relay's status endpoint. Its tests will first establish a multi-key environment using **`RACMock`** (or **`ArchiveFixtureBuilder`** for offline), then call the status endpoint with the existing request helpers and assert the credential arrays in the JSON are correct.

- **T2.c — re-anchor mechanism.** When the environment's primary ("anchor") SDK key changes, Relay must re-point its upstream connection without dropping downstream clients. These tests live mostly at the `internal/relayenv` layer, but where a full-stack check is wanted they'll use **`RACMock`** to deliver the key change and **`WithStreamRequest`** to confirm a connected SDK's stream stays alive across the re-anchor.

- **T1.c — per-key expiry and mobile-key disconnects.** Primarily `internal/relayenv`-level tests that drive the credential rotator directly, so they mostly won't need these helpers. Where a test wants to prove a *downstream* connection actually closes when a key expires, it can use **`WithStreamRequest`** and assert the stream ends.

## Notes
The lower-layer consumers (T1.c/T2.c in `internal/relayenv`) construct `EnvironmentParams` directly and don't need the RAC/archive helpers; the full-stack consumers (T3.c/T4 in `package relay`) do, and can import the `configsource` subpackage. If a future test needs true socket-level fidelity (real TCP close), a real-socket stream client can be added to `sharedtest` then; it isn't needed now.

---

> [!NOTE]
> **Low Risk**
> Test-only helpers and a reference test; no production paths or runtime behavior changes.
> 
> **Overview**
> Adds reusable **Phase 1 concurrent-keys** building blocks under `internal/sharedtest` so full-stack tests outside `package relay` can drive RAC and offline config the same way in-process relay tests already do.
> 
> A new **`internal/sharedtest/configsource`** subpackage (avoids import cycles with `envfactory`) introduces **`RACMock`** with **`MakeAutoConfigPut/Patch/DeleteEvent`**, and **`ArchiveFixtureBuilder`** to emit `.tar.gz` offline archives matching `filedata`’s metadata, data, and **`checksum.md5`** layout (including duplicate-env rejection).
> 
> **`AwaitEventOfType`** extends the existing **`WithStreamRequest`** SSE harness to wait for a specific event type with a timeout.
> 
> **`relay/concurrent_keys_harness_ref_test.go`** is a reference integration test covering offline archive → SDK stream and RAC mock → SDK stream paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9e4aba401d5dcb6e8dc786f3563a6772895b61a5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


aaron-zeisler:
FYI: I'm working to refactor this new code into `internal/sharedtest` instead of creating a new package 👍

relay/concurrent_keys_harness_ref_test.go
```diff
@@ -0,0 +1,133 @@
...
+// TestConcurrentKeysHarnessReference is the reference integration test for the Phase 1
+// concurrent-keys test helpers (SDK-2548 / T5.a). It demonstrates the reusable helpers added to
+// internal/sharedtest working together end-to-end:
+//
+//   - configsource.ArchiveFixtureBuilder (offline-mode archive with flag data)
+//   - configsource.RACMock (RAC SSE server delivering environment configuration)
+//   - sharedtest.WithStreamRequest + sharedtest.AwaitEventOfType (consuming Relay's SSE stream)
+//
+// Scenario tests for individual Phase 1 features accumulate in the sub-tasks that introduce them
+// (T1.c, T2.c, T3.c, T4), reusing these same helpers.
```
keelerm84: Lots of intermediate plan references in this block. Please reword.

keelerm84 (approved):

PR #710, opened 2026-06-18
feat: Add an expect query parameter to the status endpoints for server-side health checks

keelerm84:
Customers used /status as a health gate by fetching the JSON body,
extracting fields with jq, and comparing values in a shell. They asked
us to ship jq in the relay Docker image; a vendored binary is an attack
surface we do not control and cannot go in the distroless image.

Instead, all status endpoints (/status and the per-environment routes)
now accept a repeatable, AND-ed expect=<path><op><value> query parameter
and answer with an HTTP status code: 200 when every clause holds, 412
when a well-formed clause does not (including an absent field), and 400
when a clause is malformed. The verdict body summarizes each clause for
debugging, but the status code is the contract, so a probe or monitor
needs no body parsing and nothing installed.

The clause grammar is a small bounded path evaluator -- not an embedded
jq engine -- since the endpoints are unauthenticated. It supports map
keys, bracket-quoted keys, and array index/field-filter access, the last
of which is defined now for forward compatibility with the upcoming
concurrent-keys arrays.

---

> [!NOTE]
> **Overview**
> Status endpoints (`/status` and per-environment routes) now accept a repeatable `expect=<path><op><value>` query so probes can assert Relay health via HTTP status codes instead of parsing JSON.
> 
> Clauses are AND-ed. **200** means all held; **412** means a well-formed assertion failed (including absent optional fields or missing env keys); **422** means an unknown field or unsupported operator; **400** means the clause or query string could not be parsed. With `expect`, the body is a per-clause verdict, not the status document. Requests without `expect` are unchanged.
> 
> The evaluator is a small bounded path grammar (dotted keys, bracket-quoted keys, array index/filter for future concurrent-keys fields), validated against the status schema rather than a general query language. Query parsing fails closed so a dropped/malformed clause cannot look like a satisfied probe. Per-env **404**/**503** still win before evaluation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f5b62946cffbe1d180eac06882f193e90f995097. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


kinyoklion:
> [!NOTE]
> This comment was written by Claude (Anthropic's AI assistant), posted on Ryan Lamb's behalf as part of a multi-agent review of this PR.

**Fail-open: an `expect` clause containing a raw `;` or an invalid percent-escape is silently dropped, so the assertion is never evaluated and the endpoint returns 200.**

Both handlers read clauses via `req.URL.Query()["expect"]` (`relay/endpoints_status.go`). Go's `URL.Query()` swallows `url.ParseQuery` errors and discards the offending key/value pair — since Go 1.17 that includes any `&`-delimited segment containing a raw `;`, and any segment with a bad `%`-escape (e.g. `%zz`). The surviving segments are still returned, so:

- a **sole** malformed clause → `len(clauses) == 0` → the no-expect path: **200 + full status document**, even when the asserted condition is false;
- **one malformed clause among several** → the false clause evaporates and the response is **200 `satisfied:true`**, judged only on the survivors.

The docs promise `400 Bad Request - a clause was malformed` and tell operators to branch on the status code with no body parsing (`curl -fsS`, k8s `httpGet`), and they steer users toward hand-URL-encoding bracket-quoted environment keys — exactly where a one-character percent typo is likely. The result is a probe that reads healthy forever, silently, in the one direction a health gate must never fail.

Reproducing test — asserts the desired behavior, so it fails red on the current head (`90c71d6b`, all three subtests; also red on `872438b2`) and goes green once the query-parse error is surfaced as 400. Drop into `relay/`:

```go
package relay

import (
	"net/http"
	"testing"

	c "github.com/launchdarkly/ld-relay/v9/config"
	st "github.com/launchdarkly/ld-relay/v9/internal/sharedtest"

	"github.com/launchdarkly/go-sdk-common/v4/ldvalue"

	"github.com/stretchr/testify/assert"
)

// Query-level malformation must not silently disable assertions.
// Desired behavior per the PR contract ("400 = malformed clause", "the status code is the
// contract"): a request whose expect clause cannot even be decoded is malformed -> 400.
// It must NEVER return 200 while ignoring an assertion the caller sent.
func TestExpectQueryLevelMalformation(t *testing.T) {
	var config c.Config
	config.Environment = st.MakeEnvConfigs(st.EnvMain)

	withStartedRelay(t, config, func(p relayTestParams) {
		t.Run("raw semicolon in sole clause must not yield 200 full body", func(t *testing.T) {
			// The clause asserts status=degraded, which is FALSE (relay is healthy).
			r, _ := http.NewRequest("GET", "http://localhost/status?expect=status=degraded;x", nil)
			result, body := st.DoRequest(r, p.relay)
			t.Logf("got %d body=%s", result.StatusCode, body)
			assert.Equal(t, http.StatusBadRequest, result.StatusCode,
				"a FALSE assertion must not return 200")
		})

		t.Run("invalid percent escape in sole clause must not yield 200 full body", func(t *testing.T) {
			r, _ := http.NewRequest("GET", "http://localhost/status?expect=status%3Ddegraded%zz", nil)
			result, body := st.DoRequest(r, p.relay)
			t.Logf("got %d body=%s", result.StatusCode, body)
			assert.Equal(t, http.StatusBadRequest, result.StatusCode,
				"a FALSE assertion must not return 200")
		})

		t.Run("one droppable clause among valid ones must not yield satisfied 200", func(t *testing.T) {
			// First clause TRUE, second clause FALSE but contains a raw ';'.
			r, _ := http.NewRequest("GET",
				"http://localhost/status?expect=status=healthy&expect=status=degraded;x", nil)
			result, body := st.DoRequest(r, p.relay)
			t.Logf("got %d body=%s", result.StatusCode, body)
			assert.Equal(t, http.StatusBadRequest, result.StatusCode,
				"an assertion set containing a FALSE clause must not return 200")
			assert.False(t, ldvalue.Parse(body).GetByKey("satisfied").BoolValue())
		})
	})
}
```

Fix direction (verified green against this test with all existing tests still passing): in both `statusHandler` and `singleEnvironmentStatusHandler`, parse the raw query explicitly and fail closed —

```go
values, err := url.ParseQuery(req.URL.RawQuery)
if err != nil {
	// 400 + {"satisfied": false, "error": "malformed query string"}
	return
}
clauses := values["expect"]
```

(If you want the no-expect path to stay tolerant of unrelated malformed params, gate the 400 on the raw query containing `expect`.)

kinyoklion (approved):

PR #712, opened 2026-06-18
feat(relayenv): add ReconcileCredentials API and AcceptedSet foundation

aaron-zeisler:
## Summary

Adds the foundation for concurrent keys: an `AcceptedSet` type + builder and a new `ReconcileCredentials(newSet)` method on `EnvContext` that reconciles an environment's full accepted credential set — multiple SDK and mobile keys (each with optional per-key expiry), an anchor SDK key, a primary mobile key, and the env ID — against the current state. It also adds the anchor-only upstream-client behavior: only the anchor SDK key opens an upstream connection; other accepted keys are matched locally and served off it.

This PR is intentionally scoped as a **behavior-neutral foundation**. The existing `UpdateCredential` rotation path is retained and still drives production, and the new `ReconcileCredentials` API is not yet wired into the action handlers (it is exercised by unit tests). See [`phase1-plan.md` §2.1](.agent-docs/concurrent-keys/phase1-plan.md) for the milestone split.

Jira: [SDK-2538](https://launchdarkly.atlassian.net/browse/SDK-2538) (T1.b, under epic [SDK-2453](https://launchdarkly.atlassian.net/browse/SDK-2453)).

## Background

Phase 1 generalizes the credential model so an environment can accept a *set* of keys served off a single anchor connection. `UpdateCredential` mutates one credential kind at a time and has no notion of an anchor or a full desired state. `ReconcileCredentials` takes the complete `AcceptedSet` and owns the order of operations internally (add → set anchor → remove), so callers no longer sequence credential changes.

To keep the PR reviewable and land a working multi-key milestone before taking on rotation, this PR is limited to the data model, the API, and the anchor-only client. The pieces that change behavior or retire the legacy path are deferred:

- Wiring both handlers to `ReconcileCredentials`, removing `UpdateCredential`, building the set from the full `sdkKeys[]`/`mobileKeys[]` arrays, and payload validation → T3.c ([SDK-2547](https://launchdarkly.atlassian.net/browse/SDK-2547)).
- Per-key grace-period deprecation + the cleanup ticker and downstream disconnect → T1.c ([SDK-2539](https://launchdarkly.atlassian.net/browse/SDK-2539)).
- The robust re-anchor swap and the re-queue-on-primary-switch fixes (SDK + mobile) → T2.c ([SDK-2542](https://launchdarkly.atlassian.net/browse/SDK-2542)).

### Anchor / primary designations and malformed sets

`AcceptedSet` carries its own anchor (the SDK key that owns the upstream connection) and primary mobile key, both designated through the builder via `WithPrimarySDKKey` / `WithPrimaryMobileKey` (which add the key and designate it in one call; the builder de-duplicates by value). `Build()` requires an anchor and returns a structured `*MalformedCredentialSetError` if none was designated; `Reconcile` guards against a set with no anchor and makes no state change in that case. Structural validation of the wire payload (undefined credentials, an anchor absent from the array) is handled upstream when the payload is parsed — see T3.c.

## Changes

- **`internal/credential/accepted_set.go`** (new) — the `AcceptedSet` type: the accepted SDK/mobile keys (each with optional per-key expiry), anchor SDK key, primary mobile key, and env ID, plus `MalformedCredentialSetError`.
- **`internal/credential/accepted_set_builder.go`** (new) — `AcceptedSetBuilder` (`WithSDKKey` / `WithExpiringSDKKey` / `WithPrimarySDKKey` / `WithMobileKey` / `WithExpiringMobileKey` / `WithPrimaryMobileKey` / `WithEnvironmentID` / `Build`, de-duping by value). Split from the type so the data model and its construction live in separate files.
- **`internal/credential/rotator.go`** — `Rotator.Reconcile(set, now)` does add / set-anchor / remove against the accepted-set maps, storing per-key expiry as data. The SDK and mobile diffs share a single generic `reconcileAcceptedKeys` helper (parameterized over the key type) so the algorithm lives in one place. The legacy `Rotate` / `RotateWithGrace` mutators also keep those maps in sync, so `PrimaryCredentials` is correct on both the reconcile and legacy paths.
- **`internal/relayenv/env_context.go` / `env_context_impl.go`** — the `ReconcileCredentials` interface method + implementation (with a time-injectable seam for deterministic tests), and the anchor-only gate in `addCredential` (only the anchor opens an upstream SDK client and repoints event/metrics forwarding). `UpdateCredential` / `CredentialUpdate` are retained.
- **`relay/autoconfig_actions.go`** — keeps a nil-env `return` guard in `AddEnvironment`; otherwise the handlers are unchanged and continue to use `UpdateCredential`.
- **Tests** — unit coverage split to mirror the source: builder validation/de-dup (`accepted_set_builder_test.go`), the `MalformedCredentialSetError` message (`accepted_set_test.go`), and the `Reconcile` behavior — multi-key accept/route, the anchor-only client, and the no-anchor guard (`rotator_test.go`). Auto-config expiring-key tests assert a single anchor client.

Targets `feat/concurrent-keys` (not the default branch).

[SDK-2538]: https://launchdarkly.atlassian.net/browse/SDK-2538?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SDK-2547]: https://launchdarkly.atlassian.net/browse/SDK-2547?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Touches credential acceptance, primary credential listing, and upstream client/event wiring, but the new reconcile API is not yet used by auto-config handlers; legacy rotation behavior is adjusted to match accepted-set maps.
> 
> **Overview**
> Introduces **`AcceptedSet`** and **`AcceptedSetBuilder`** so an environment can describe a full desired credential state (multiple SDK/mobile keys with optional expiry, anchor SDK key, primary mobile key, env ID), with **`Build()`** rejecting empty sets and sets without a designated anchor.
> 
> Adds **`Rotator.Reconcile`** and **`EnvContext.ReconcileCredentials`** to diff that desired set against current state, queue additions/expirations, and apply them via the existing **`triggerCredentialChanges`** path. **`PrimaryCredentials`** now reflects all accepted non-deprecated keys in the rotator maps; the legacy **`Rotate` / `RotateWithGrace`** path is updated to keep those maps aligned (including revoking the previous key when SDK grace is already expired).
> 
> **Anchor-only upstream behavior:** in **`addCredential`**, only the anchor SDK key starts an upstream client and repoints SDK event/metrics forwarding; only the **primary** mobile key repoints mobile event forwarding. Extra accepted keys still get streams, handlers, and connection mappings.
> 
> Auto-config **`AddEnvironment`** returns early when env init fails so it does not call **`UpdateCredential`** on a nil env. Production handlers still use **`UpdateCredential`**; **`ReconcileCredentials`** is exercised in tests only.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e5f94e7bbe5c829c23797ab6c0ceb992d5f1a3b0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/relayenv/env_context.go
```diff
@@ -21,41 +21,130 @@ import (
...
-	c.now = t
-	return c
+// AcceptedSet is the full set of credentials that an environment should accept after a
+// reconcile. It carries every accepted server-side SDK key and mobile key — each with an
+// optional per-key expiry — plus the single environment ID. The anchor (the one SDK key that
+// owns the environment's upstream connection) is supplied separately to ReconcileCredentials.
+//
+// Per design §6.3 ("trust the array"), a key's expiry is taken from its entry in this set; the
+// legacy sdkKey.expiring{} wire slot is not consulted when building it.
+//
```
aaron-zeisler: Remove the reference to the design document from this comment.

internal/relayenv/env_context.go
```diff
@@ -21,41 +21,130 @@ import (
...
+}
+
+// WithMobileKey adds a permanent (non-expiring) mobile key to the set. It is a no-op if the key
+// is undefined.
+func (s AcceptedSet) WithMobileKey(key config.MobileKey) AcceptedSet {
+	if key.Defined() {
+		s.mobileKeys = append(s.mobileKeys, acceptedMobileKey{key: key})
+	}
+	return s
+}
```
aaron-zeisler, line 42: Is there a need for a `WithExpiringMobileKey()` function?

internal/relayenv/env_context.go
```diff
@@ -21,41 +21,130 @@ import (
...
+		if k.key == key {
+			return true
+		}
+	}
+	return false
+}
+
+// deprecatedSDKKey returns the accepted SDK key other than the anchor that carries an expiry —
+// the key being phased out with a grace period. Phase 1's single-key model has at most one such
+// key; multi-key reconciliation is generalized in T1.c.
```
aaron-zeisler: Remove the reference to "phase 1" from this comment - this is a direct reference to a specific project.

internal/relayenv/env_context.go
```diff
@@ -21,41 +21,130 @@ import (
...
+		if k.key != anchor && k.expiry != nil {
+			return k, true
+		}
+	}
+	return acceptedSDKKey{}, false
+}
+
+// primaryMobileKey returns the environment's primary mobile key — the single accepted mobile key
+// in Phase 1. It returns false if the set has no mobile key. Per-key mobile expiry and multi-key
+// mobile cleanup are generalized in T1.c.
```
aaron-zeisler: Remove the reference to "phase 1" and the specific task T1.c - these are direct references to a specific project, which we don't want in the permanent code.

internal/relayenv/env_context.go
```diff
@@ -21,41 +21,130 @@ import (
...
+
+// MalformedCredentialSetError is returned by ReconcileCredentials when the supplied anchor SDK
+// key is not present among the accepted set's SDK keys — a violation of the backend invariant
+// that the anchor (sdkKey.value) always appears in sdkKeys[] (design §4.2, §9).
+//
+// When ReconcileCredentials returns this error it has made no changes, so the environment's
+// previous accepted set is preserved. The caller is responsible for the second half of the
+// malformed-payload policy: reconnecting the RAC stream with jitter to force a fresh put (wired
+// in T3.c). RAC is one-way push with no NAK channel, so without the reconnect the backend would
+// believe the malformed patch was applied and would not send fresh state.
```
aaron-zeisler: Remove the project-specific references from this comment and all other comments throughout the pull request.

internal/relayenv/env_context.go
```diff
@@ -21,41 +21,130 @@ import (
...
+
+type acceptedSDKKey struct {
+	key    config.SDKKey
+	expiry *time.Time // nil = permanent
+}
+
+type acceptedMobileKey struct {
+	key    config.MobileKey
+	expiry *time.Time // nil = permanent
+}
```
aaron-zeisler: These acceptedSDKKey and acceptedMobileKey structs are very similar to the structs introduced in https://github.com/launchdarkly/ld-relay/pull/702. Is there an opportunity for reuse? Or is it better to leave them separated?

relay/autoconfig_actions.go
```diff
@@ -50,15 +66,10 @@ func (a *relayAutoConfigActions) UpdateEnvironment(params envfactory.Environment
...
 
-	if params.MobileKey.Defined() {
-		env.UpdateCredential(relayenv.NewCredentialUpdate(params.MobileKey))
-	}
 	if params.SDKKey.Defined() {
-		update := relayenv.NewCredentialUpdate(params.SDKKey)
-		if params.ExpiringSDKKey.Defined() {
-			update = update.WithGracePeriod(params.ExpiringSDKKey.Key, params.ExpiringSDKKey.Expiration)
+		if err := env.ReconcileCredentials(acceptedSetFromParams(params), params.SDKKey); err != nil {
+			a.r.loggers.Errorf(logMsgReconcileCredentialsError, params.Identifiers.GetDisplayName(), err)
```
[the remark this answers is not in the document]
aaron-zeisler: @keelerm84  Here Bugbot's describing a situation where we got a payload that contained a mobile key but no SDK key. Do you think that's in the realm of possibilities? Wondering if we should handle that situation at all, either gracefully or by rejecting the payload.

internal/relayenv/env_context.go
aaron-zeisler: Keeping them separate.

internal/relayenv/env_context.go
```diff
@@ -21,41 +21,116 @@ import (
...
+	return s
+}
+
+// hasSDKKey reports whether key is one of the set's accepted SDK keys.
+func (s AcceptedSet) hasSDKKey(key config.SDKKey) bool {
+	for _, k := range s.sdkKeys {
+		if k.key == key {
+			return true
+		}
+	}
```
aaron-zeisler: Curiosity question: Is there a precedent within this repository, or would it be beneficial, to use a generic helper/utility function for these types of operations against a slice? I'm thinking about the built-in `slices.Contains()` function.

internal/relayenv/env_context.go
```diff
@@ -21,41 +21,116 @@ import (
...
+}
+
+// deprecatedSDKKey returns the accepted SDK key other than the anchor that carries an expiry —
+// the key being phased out with a grace period. There is at most one such key today.
+func (s AcceptedSet) deprecatedSDKKey(anchor config.SDKKey) (acceptedSDKKey, bool) {
+	for _, k := range s.sdkKeys {
+		if k.key != anchor && k.expiry != nil {
+			return k, true
+		}
+	}
```
aaron-zeisler: Similar comment re: slice utility methods

internal/relayenv/env_context.go
```diff
@@ -21,41 +21,116 @@ import (
...
+	// Anchor is the anchor credential that was not found among the set's SDK keys.
+	Anchor credential.SDKCredential
+}
+
+func (e *MalformedCredentialSetError) Error() string {
+	if e.Anchor == nil {
+		return "malformed credential set: anchor SDK key is missing"
+	}
+	return fmt.Sprintf("malformed credential set: anchor SDK key %s is not present in the accepted set",
+		e.Anchor.Masked())
```
aaron-zeisler: Question: here we're logging the SDK key's credential. In [this pull request comment thread](https://github.com/launchdarkly/ld-relay/pull/702#discussion_r3431066362), we found that it's possible for the SDK Key's `key` value to be empty in an edge case. In that edge case, will an empty string be written to the log here?

internal/relayenv/env_context.go
```diff
@@ -21,41 +22,114 @@ import (
...
+// AcceptedSet is the full set of credentials that an environment should accept after a
+// reconcile. It carries every accepted server-side SDK key (each with an optional expiry) and
+// mobile key, plus the single environment ID. The anchor (the one SDK key that owns the
+// environment's upstream connection) is supplied separately to ReconcileCredentials.
+//
+// A key's expiry is taken from its entry in this set; the legacy sdkKey.expiring{} wire slot is
+// not consulted when building it.
+//
+// Build an AcceptedSet with NewAcceptedSet and the With* methods, which ignore undefined
+// credentials so callers can supply optional keys unconditionally.
```
keelerm84: I'm not sure how I feel about that decision. I would be concerned that it would lead to a silent failure for a customer that would be difficult to track down. What do you think?

internal/relayenv/env_context.go
```diff
@@ -21,41 +22,114 @@ import (
...
+	}
+	return s
+}
+
+// WithExpiringSDKKey adds an SDK key that should be accepted until the given expiry. It is a
+// no-op if the key is undefined.
+func (s AcceptedSet) WithExpiringSDKKey(key config.SDKKey, expiry time.Time) AcceptedSet {
+	if key.Defined() {
+		exp := expiry
+		s.sdkKeys = append(s.sdkKeys, acceptedSDKKey{key: key, expiry: &exp})
```
keelerm84: Is this not valid?

```suggestion
		s.sdkKeys = append(s.sdkKeys, acceptedSDKKey{key: key, expiry: &expiry})
```

internal/relayenv/env_context.go
```diff
@@ -21,41 +22,114 @@ import (
...
+}
+
+// NewAcceptedSet returns an empty AcceptedSet.
+func NewAcceptedSet() AcceptedSet {
+	return AcceptedSet{}
+}
+
+// WithSDKKey adds a permanent (non-expiring) SDK key to the set. It is a no-op if the key is
+// undefined.
+func (s AcceptedSet) WithSDKKey(key config.SDKKey) AcceptedSet {
```
keelerm84: The API seems to suggest that operating on an existing `AcceptedSet` (value receiver) wouldn't affect it. Rather, we would return a new copy that includes your changes. 

For example, 

```go
set1 := NewAcceptedSet().WithSDKKey("key1")
set2 := set1.WithSDKKey("key2")
```

both sets now have `key1` and `key2`.

I'm wondering if we don't want to separate the idea of the type from a builder, and have the builder operate on pointer values so it's more clear you are mutating the underlying instance. The builder also gives us a nice place to put any final validation, like ensuring at least one server key has been provided.

Thoughts?

internal/relayenv/env_context.go
```diff
@@ -21,41 +22,114 @@ import (
...
+		return k.key != anchor && k.expiry != nil
+	})
+	if i < 0 {
+		return acceptedSDKKey{}, false
+	}
+	return s.sdkKeys[i], true
+}
+
+// primaryMobileKey returns the environment's single accepted mobile key, or false if there is none.
+func (s AcceptedSet) primaryMobileKey() (config.MobileKey, bool) {
```
keelerm84: Is the idea that this will be expanding to support multiple keys in a future PR?

internal/relayenv/env_context.go
```diff
@@ -21,41 +22,114 @@ import (
...
-// WithTime overrides the update's current time for testing purposes.
-// Because the environment's credential rotation algorithm compares the current time to the specific expiry of
-// each credential, this can be used to trigger behavior in a more predictable way than relying on the actual time
-// in the test.
-func (c *CredentialUpdate) WithTime(t time.Time) *CredentialUpdate {
-	c.now = t
-	return c
+// deprecatedSDKKey returns the accepted SDK key other than the anchor that carries an expiry —
+// the key being phased out with a grace period. There is at most one such key today.
+func (s AcceptedSet) deprecatedSDKKey(anchor config.SDKKey) (acceptedSDKKey, bool) {
```
keelerm84: Can't there be more than one?

internal/relayenv/env_context_impl.go
```diff
@@ -550,13 +550,45 @@ func (c *envContextImpl) SetIdentifiers(ei EnvIdentifiers) {
...
+		// Malformed payload: make no changes (preserve the previous accepted set) and surface a
+		// structured error so the caller can reconnect the RAC stream.
+		return &MalformedCredentialSetError{Anchor: anchor}
+	}
+
+	// Re-anchor: the anchor becomes the primary SDK key. A non-anchor SDK key carrying an expiry is
+	// a graceful deprecation, and its expiry is read from the set rather than the legacy expiring
+	// slot. The dedicated upstream-client swap is handled separately; here the swap is realized by
+	// the rotation below, which starts a client for the new anchor via addCredential.
+	if deprecated, hasDeprecated := newSet.deprecatedSDKKey(anchorKey); hasDeprecated {
```
keelerm84: Mentioned above, but this looks like it is assuming you can only have 2 keys going on at the same time (current restriction). Is this being expanded in a subsequent PR?

relay/autoconfig_actions.go
```diff
@@ -13,8 +13,22 @@ const (
...
 )
 
+// acceptedSetFromParams builds the full accepted credential set from an environment's parameters,
+// using the singular SDK/mobile/env fields plus the optional expiring SDK key.
+func acceptedSetFromParams(params envfactory.EnvironmentParams) relayenv.AcceptedSet {
+	set := relayenv.NewAcceptedSet().
+		WithSDKKey(params.SDKKey).
+		WithMobileKey(params.MobileKey).
+		WithEnvironmentID(params.EnvID)
+	if params.ExpiringSDKKey.Defined() {
```
keelerm84: All of these other methods are handling undefined values automatically but not here. Why not make it consistent?

relay/autoconfig_actions.go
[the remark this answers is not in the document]
keelerm84: I don't think you can run the relay without providing a server key. Technically it probably wouldn't need one when operating in offline mode. But the config certainly includes one since it mirrors RAC so I think it's reasonable to just enforce it going forward.

keelerm84 (commented):
While reviewing this, I uncovered a nil panic case. Pushed a small commit to just piggy back on this PR. LMK if you'd rather we split it out, but it's such a small line I think it's probably fine.

internal/relayenv/env_context.go
aaron-zeisler: lol thank you

internal/relayenv/env_context.go
aaron-zeisler: Yeah, I agree that both returning the AcceptedSet and mutating it is a weird API. But I think that this is actually common for the Builder pattern in Go. I'll look at the other builder/constructor implementations in Relay proxy for inspiration.


> have the builder operate on pointer values so it's more clear you are mutating the underlying instance

Yes, I think this is the key 👍

aaron-zeisler: I've refactored the builder. Now there's a concrete AcceptedSetBuilder object. It has several "With..." functions, and it has a Build() function that returns the AcceptedSet object.

relay/autoconfig_actions.go
aaron-zeisler: Done 👍

relay/autoconfig_actions.go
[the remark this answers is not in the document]
aaron-zeisler: Thanks. The builder now verifies that at least one SDK key is provided 👍

internal/relayenv/env_context.go
aaron-zeisler: I'm going to refactor this code and introduce more logic that processes multiple keys now. These `primaryMobileKey()` and `deprecatedSDKKey()` functions are basically temporary or throw-away, since they'll be replaced evolved later. I don't think there's any use in writing them if I'm just going to trash them later (like, tomorrow probably).

Note to self: after getting this feature working end-to-end, I'll want to perform a review on all the code to understand if there's duplication similar to this, or if we can throw away code that's either no longer used, or code that ended up being temporary to this project. Lesson learned: breaking down the project into the small chunks that I've defined is confusing the agent or making it perform unnecessary work, and it would have been better to make slightly larger tasks. The trade-off is reviewability: so far we've had really good conversations in these PRs, I think because they're so small that we can focus on the details without being overwhelmed by the overall size of the change.

internal/credential/accepted_set.go
```diff
@@ -0,0 +1,133 @@
...
+
+// hasSDKKey reports whether key is one of the set's accepted SDK keys.
+func (s AcceptedSet) hasSDKKey(key config.SDKKey) bool {
+	for _, k := range s.sdkKeys {
+		if k.key == key {
+			return true
+		}
+	}
+	return false
+}
```
aaron-zeisler, line 44: Can this function use `slices.Contains()` or another helper function from Go's built-in slices package?

internal/credential/accepted_set.go
```diff
@@ -0,0 +1,133 @@
...
+
+// WithExpiringMobileKey adds a mobile key that should be accepted until the given expiry. It is a
+// no-op if the key is undefined.
+func (b *AcceptedSetBuilder) WithExpiringMobileKey(key config.MobileKey, expiry time.Time) *AcceptedSetBuilder {
+	if key.Defined() {
+		b.set.mobileKeys = append(b.set.mobileKeys, acceptedMobileKey{key: key, expiry: &expiry})
+	}
+	return b
+}
```
aaron-zeisler: The previous 4 functions (WithMobileKey, WithExpiringMobileKey, with SDKKey, WithExpiringSDKKey) all take a single key. Would it be beneficial if these functions took an array of keys and their expirations instead?

Another question: would it be beneficial (or just unnecessarily complex) to have the `expiry` parameter be optional? So we could reduce the 4 functions down to 2. I think this would require using Go's functional options pattern.

internal/credential/rotator.go
```diff
@@ -116,14 +115,44 @@ func (r *Rotator) PrimaryCredentials() []SDKCredential {
...
-	return slices.DeleteFunc([]SDKCredential{
-		r.primarySdkKey,
-		r.primaryMobileKey,
-		r.primaryEnvironmentID,
-	}, func(cred SDKCredential) bool {
-		return !cred.Defined()
-	})
+	creds := make([]SDKCredential, 0, len(r.acceptedSDKKeys)+len(r.acceptedMobileKeys)+1)
+
+	// Anchor first, then any other accepted (non-deprecated) SDK keys.
```
aaron-zeisler: Is it necessary to add the primary/anchor key first? If order doesn't matter, then we could get rid of this if block, plus the `if key == r.primarySdkKey` check inside the for block just below.

Same question for primaryMobileKey below.

internal/credential/rotator.go
```diff
@@ -327,3 +357,145 @@ func (r *Rotator) StepTime(now time.Time) (additions []SDKCredential, expiration
...
+			continue
+		}
+		delete(r.acceptedMobileKeys, key)
+		r.expirations = append(r.expirations, key)
+		r.loggers.Infof("Mobile key %s is no longer accepted and has been revoked", key.Masked())
+	}
+
+	// The primary mobile key — used where a single mobile key is required (e.g. event forwarding) —
+	// is the first permanent key in the set, falling back to the first non-expired key, else empty.
+	r.primaryMobileKey = ""
```
aaron-zeisler: Per the design document, aren't we using the single mobileKey that's defined in the RAC payload as the primary one? Or did we make an explicit decision not to use that one and instead use the first permanent key in the set?

internal/relayenv/env_context.go
aaron-zeisler: Thanks for this comment. I've decided to add validation so undefined credentials are caught when the RAC payload is received and parsed and processed (that's when this AcceptedSet struct is populated and built). That validation code will come in a subsequent PR (for the SDK-2534 ticket, I think).

internal/credential/accepted_set.go
```diff
@@ -0,0 +1,170 @@
...
+
+// WithAnchorSDKKey designates key as the set's anchor — the SDK key that owns the environment's
+// upstream connection. The anchor must also be added to the set (via WithSDKKey or
+// WithExpiringSDKKey); Build verifies this. It is a no-op if the key is undefined.
+func (b *AcceptedSetBuilder) WithAnchorSDKKey(key config.SDKKey) *AcceptedSetBuilder {
+	if key.Defined() {
+		b.set.primarySdkKey = key
+	}
+	return b
+}
```
aaron-zeisler: A couple ideas to simplify this interface and its logic:
* Rename this function to WithPrimarySDKKey
* Should WithPrimarySdkKey and WithPrimaryMobileKey set the primarySdkKey and primaryMobileKey and ALSO add the key to the sdkKeys / mobileKeys arrays? The interface is currently a bit clunky, since callers need to call 2 "With..." methods when adding some keys.
* What if WithSDKKey and WithMobileKey had another input parameter "isPrimary" that was a boolean? And then the function additionally set the primary***Key variable when the input parameter was true.

General question: Is there any validation, prevention, or de-duping necessary if the caller has added the same key twice.

internal/credential/accepted_set.go
```diff
@@ -0,0 +1,87 @@
...
+// single key takes one call. Build requires that an anchor was designated. (Structural validation of
+// the wire payload — undefined credentials, an anchor absent from the array — happens upstream when
+// the payload is parsed into the set; see SDK-2547.)
+//
+// A key's expiry is taken from its entry in this set; the legacy sdkKey.expiring{} wire slot is not
+// consulted when building it.
+//
+// Construct an AcceptedSet with AcceptedSetBuilder (see accepted_set_builder.go).
+type AcceptedSet struct {
+	sdkKeys          []acceptedSDKKey
```
keelerm84: Have you considered having the `AcceptedSet` actually store values as a set? This handles dedupes, and removes the need for the `ContainsFunc` searches below.

internal/credential/rotator.go
```diff
@@ -327,3 +358,127 @@ func (r *Rotator) StepTime(now time.Time) (additions []SDKCredential, expiration
...
+// It returns a *MalformedCredentialSetError, without mutating any state, if the set's anchor is
+// undefined or not one of its SDK keys. AcceptedSetBuilder.Build already enforces this, so a set
+// obtained from the builder always reconciles cleanly; the guard defends against a zero-value set.
+func (r *Rotator) Reconcile(set AcceptedSet, now time.Time) error {
+	if !set.primarySdkKey.Defined() {
+		return &MalformedCredentialSetError{Anchor: nil}
+	}
+	if !set.hasSDKKey(set.primarySdkKey) {
+		return &MalformedCredentialSetError{Anchor: set.primarySdkKey}
+	}
```
keelerm84: Do we need to do these checks here? Aren't they now enforced by the builder?

internal/credential/accepted_set.go
aaron-zeisler: Ha! Of course, that makes perfect sense. I'll do it, thank you for the recommendation.

aaron-zeisler: Interestingly, Go doesn't have a built-in Set type. There are a couple open-source options (including one from Hashicorp), and there's one defined in our Foundation repository, but I decided to go with a map to get the same benefits and keep it simple. If you feel strongly against the map, I could import the Hashicorp library or roll my own (similar to the Foundation code).

keelerm84: No, I was expecting you would use a map. Sorry, I should have been more clear on that.

aaron-zeisler: No worries! I'm used to working in an environment with access to an actual "Set" utility class, so I was a surprised by the lack of one.

keelerm84 (approved):

PR #714, opened 2026-06-19
feat(credential): generalize cleanup ticker for per-key expiry + mobile-key disconnects

aaron-zeisler:
## Summary

Generalizes `Rotator.StepTime` to enforce per-key expiry for **mobile keys** and for the **reconcile path**, symmetric with the existing SDK/legacy grace path. Completes T1.c of the concurrent-keys plan. `DeprecatedCredentials` now surfaces deprecated mobile keys alongside deprecated SDK keys. The downstream-disconnect path (`triggerCredentialChanges` → `removeCredential` → `envStreams.RemoveCredential`) already handles any credential kind, so no production changes to `env_context_impl` are required beyond new tests.

[Jira: SDK-2539](https://launchdarkly.atlassian.net/browse/SDK-2539)

## Background

The `Rotator` foundation (SDK-2538, #712) stores per-key expiry as data on each accepted-set entry (`acceptedKeyInfo.expiry`) and populates `deprecatedMobileKeys` via the legacy `RotateWithGrace` path — but nothing acted on either: `StepTime` only walked `deprecatedSdkKeys`, there was no `expireMobileKey`, and `deprecatedCredentials()` was SDK-only. This PR closes that gap so the cleanup ticker enforces expiry uniformly across both key kinds and both expiry mechanisms.

## StepTime design decision

There were two ways to make `StepTime` act on reconcile-path expiry: (a) read `acceptedKeyInfo.expiry` directly off the accepted-set entry, or (b) first migrate expiring reconcile-path keys into the deprecated maps and let the existing deprecated-map walk drop them.

This PR takes **(a)**. `StepTime` expires reconcile-path keys in place by reading the expiry stored on the accepted entry, and the deprecated maps remain exclusively the legacy `RotateWithGrace` grace mechanism (which `StepTime` continues to drain for both SDK and mobile keys). This keeps #712's "per-key expiry is data on the accepted entry" model intact and is symmetric for both key kinds.

Consequence (consistent with #712): a reconcile-path key with a future expiry appears in `PrimaryCredentials` (accepted, non-deprecated) until its expiry elapses, whereas a legacy grace-period key appears in `DeprecatedCredentials` during its grace window. Expiry is strict (`now` must be strictly after the expiry timestamp) across all four `StepTime` loops, preserving the established SDK-key semantics.

The re-anchor swap and mobile-primary-switch re-queue are intentionally **out of scope** here (T2.c / SDK-2542).

## Changes

- `StepTime` now enforces expiry across four loops: the legacy `deprecatedSdkKeys` / `deprecatedMobileKeys` grace maps, and the reconcile-path `acceptedSDKKeys` / `acceptedMobileKeys` maps keyed by `acceptedKeyInfo.expiry` (a `nil` expiry is permanent and never dropped).
- `expireMobileKey` helper added, mirroring `expireSDKKey` — deletes from both `deprecatedMobileKeys` **and** `acceptedMobileKeys` (the latter is load-bearing: `PrimaryCredentials` derives from `acceptedMobileKeys`, so an expired key would otherwise linger).
- `deprecatedCredentials()` now includes entries from `deprecatedMobileKeys`.
- Comments on `deprecatedMobileKeys`, `updateMobileKey`, and `Reconcile` updated to reflect that the cleanup ticker now acts on the stored expiry.
- New rotator tests: `TestReconcileExpiringKeysAreEvictedByStepTime` (reconcile path, SDK + mobile, end-to-end), `TestMobileKeyDeprecation`, `TestManyConcurrentMobileKeyDeprecation`, `TestMixedSDKAndMobileKeyExpiry`, `TestDeprecatedCredentialsIncludesMobileKeys`. `TestRotateWithGraceMobileKey/does-not-panic` extended with a post-expiry assertion; `TestReconcileAcceptsExpiringKeysAsData` comment corrected.
- New env-level tests: `TestMobileKeyGraceExpiry` (legacy path) and `TestMobileKeyReconcileExpiry` (reconcile path), both driving a mobile key to expiry through `triggerCredentialChanges`.

### DIFF
--- a/internal/credential/rotator.go
+++ b/internal/credential/rotator.go
@@ -20,8 +20,8 @@ type Rotator struct {
 	primaryMobileKey config.MobileKey
 
 	// deprecatedMobileKeys stores mobile keys being phased out with a grace period, keyed
-	// by credential value with the associated expiry time. StepTime does not yet act on
-	// these entries; generalizing the cleanup ticker to drop them is handled separately.
+	// by credential value with the associated expiry time. StepTime walks this map and emits
+	// an expiration once a key's grace period passes, mirroring deprecatedSdkKeys.
 	deprecatedMobileKeys map[config.MobileKey]time.Time
 
 	// There is only one environment ID active at a given time, and it won't actually be rotated. The mechanism is
@@ -141,8 +141,11 @@ func (r *Rotator) primaryCredentials() []SDKCredential {
 }
 
 func (r *Rotator) deprecatedCredentials() []SDKCredential {
-	deprecated := make([]SDKCredential, 0, len(r.deprecatedSdkKeys))
+	deprecated := make([]SDKCredential, 0, len(r.deprecatedSdkKeys)+len(r.deprecatedMobileKeys))
 	for key := range r.deprecatedSdkKeys {
+		deprecated = append(deprecated, key)
+	}
+	for key := range r.deprecatedMobileKeys {
 		deprecated = append(deprecated, key)
 	}
 	return deprecated
@@ -225,8 +228,8 @@ func (r *Rotator) updateEnvironmentID(envID config.EnvironmentID) {
 }
 
 // updateMobileKey sets a new primary mobile key. When grace is nil the outgoing key is
-// immediately revoked; when non-nil its expiry is stored in deprecatedMobileKeys.
-// StepTime does not yet act on deprecatedMobileKeys; that cleanup is handled separately.
+// immediately revoked; when non-nil its expiry is stored in deprecatedMobileKeys for the
+// cleanup ticker (StepTime) to act on.
 func (r *Rotator) updateMobileKey(mobileKey config.MobileKey, grace *GracePeriod) {
 	r.mu.Lock()
 	defer r.mu.Unlock()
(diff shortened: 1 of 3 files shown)


internal/credential/rotator.go
```diff
@@ -322,6 +331,12 @@ func (r *Rotator) StepTime(now time.Time) (additions []SDKCredential, expiration
 		}
 	}
 
+	for key, expiry := range r.deprecatedMobileKeys {
+		if now.After(expiry) {
+			r.expireMobileKey(key)
+		}
+	}
+
```
aaron-zeisler, line 382: Curiosity question: Throughout this pull request, an extra loop or an extra function is necessary to process mobile keys as opposed to SDK keys. Are there advantages of keeping separate arrays/functions for sdk keys vs mobile keys? Or could we combine some of these arrays and functions?

I'm specifically talking about:
- deprecatedMobileKeys vs deprecatedSdkKeys, and the multiple loops required to process them
- expireMobileKey() vs expireSdkKey()

aaron-zeisler: From Claude's code review:

Good question. The expiry/cleanup path on its own is mergeable — `additions` and `expirations` are already unified as `[]SDKCredential`, and both `config.SDKKey` and `config.MobileKey` satisfy the `credential.SDKCredential` interface, so the two deprecated maps could in principle collapse into one `map[credential.SDKCredential]time.Time` with a single `expire()` helper and one loop in `StepTime`.

I kept them separate for three reasons:

1. **The SDK-key path carries logic the mobile path doesn't.** `updateSDKKey` does stale-deprecation detection keyed on `grace.key` (the `deprecatedSdkKeys[grace.key]` lookup), handling the case where the upstream re-sends an already-seen `deprecated` field. Mobile keys have no analog. The typed `map[config.SDKKey]` keeps that lookup compile-time safe — a merged interface-keyed map would let `deprecatedSdkKeys[someMobileKey]` compile, which is exactly the mismatch the separate types prevent.

2. **It's not just these two maps.** The accepted-set tracking one level up is split the same way (`acceptedSDKKeys` vs `acceptedMobileKeys`). Collapsing only the deprecated maps would leave the rotator half-merged and internally inconsistent; doing it properly means unifying the accepted-set maps too.

3. **There's a cleaner home for the unification.** The model is heading toward a single keyed "accepted set + anchor" structure that holds server and mobile keys together. That's the right place to collapse the SDK/mobile duplication in one coherent move, rather than piecemeal inside the cleanup ticker.

A generics helper (`func stepExpiry[K ...](m map[K]time.Time, ...)`) could dedupe the two loops without merging the maps, but for two call sites it adds more indirection than the handful of duplicated lines it removes.

So: leaving the split as-is here intentionally, with the broader server+mobile unification handled where the keyed-set model is consolidated.

keelerm84 (approved):

PR #715, opened 2026-06-19
feat(relayenv): gate startSDKClient on anchor key

aaron-zeisler:
## Summary
Non-anchor server SDK keys now flow through `addCredential` and receive `envStreams` registration, handler bundles, and connection-mapper mapping — but `startSDKClient` is only called when the added key is the current anchor. This is design §4.1 applied: one upstream connection per environment, on the anchor key; all other accepted server keys match locally.

The event dispatcher and metrics publisher are likewise only updated on anchor addition (design §4.3: events collapse to the anchor per kind).

[Jira: SDK-2540](https://launchdarkly.atlassian.net/browse/SDK-2540)

## Background
Previously `addCredential` called `startSDKClient` for every `config.SDKKey` it received. As long as every env had exactly one SDK key this was harmless — that key was always the anchor. Phase 1 (concurrent keys) breaks that assumption: a reconcile now delivers additional non-anchor SDK keys through `addCredential`, and each of them must not open its own upstream connection.

To route non-anchor keys through `addCredential`, `reconcileCredentials` now calls `Rotator.AddNonAnchorSDKKey` for each non-anchor entry in the accepted set. That method enqueues the key in the rotator's additions slice (with an idempotency guard via `acceptedSDKKeys`) so it is delivered by `triggerCredentialChanges` while preserving the add-before-remove ordering guarantee.

`expireSDKKey` is also updated to remove the expired key from `acceptedSDKKeys`, so a subsequent `AddNonAnchorSDKKey` call for a revoked-then-re-added key is treated as new rather than a no-op.

## Changes
- **`rotator.go`**: add `AddNonAnchorSDKKey`; clean up `acceptedSDKKeys` in `expireSDKKey`
- **`env_context_impl.go`**: gate `startSDKClient` + `ReplaceCredential` in `addCredential` on `key == c.keyRotator.SDKKey()`; call `AddNonAnchorSDKKey` for non-anchor entries in `reconcileCredentials`
- **`env_context_impl_test.go`**: `TestNonAnchorSDKKeysDoNotOpenUpstreamClient` — anchor + 2 non-anchor SDK keys, assert only one upstream client opened

### DIFF
--- a/internal/relayenv/env_context_impl.go
+++ b/internal/relayenv/env_context_impl.go
@@ -437,11 +437,11 @@ func (c *envContextImpl) addCredential(newCredential credential.SDKCredential) {
 	}
 
 	// A new SDK key means:
-	//  1. we should start a new SDK client*, but only for the anchor (design §4.1: one upstream
-	//     connection per env, on the anchor). Non-anchor server keys get envStreams + handler bundles
-	//     above, but no upstream client — matching today's mobile-key behavior.
+	//  1. we should start a new SDK client*, but only for the anchor: there is a single upstream
+	//     connection per environment, owned by the anchor key. Non-anchor server keys get envStreams
+	//     + handler bundles above, but no upstream client — matching today's mobile-key behavior.
 	//  2. we should tell all event forwarding components that use an SDK key to use the new one,
-	//     again only when it is the anchor (events collapse to the anchor per kind — design §4.3).
+	//     again only when it is the anchor, since events collapse to the anchor per kind.
 	// A new mobile key does not require starting a new SDK client, but does requiring updating any event forwarding
 	// components that use a mobile key.
 	// *Note: we only start a new SDK client in online mode. This is somewhat of an architectural hack because EnvContextImpl
--- a/internal/relayenv/env_context_impl.go
+++ b/internal/relayenv/env_context_impl.go
@@ -437,8 +437,11 @@ func (c *envContextImpl) addCredential(newCredential credential.SDKCredential) {
 	}
 
 	// A new SDK key means:
-	//  1. we should start a new SDK client*
-	//  2. we should tell all event forwarding components that use an SDK key to use the new one.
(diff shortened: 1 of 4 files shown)


internal/relayenv/env_context_impl.go
```diff
@@ -437,8 +437,11 @@ func (c *envContextImpl) addCredential(newCredential credential.SDKCredential) {
 	}
 
 	// A new SDK key means:
-	//  1. we should start a new SDK client*
-	//  2. we should tell all event forwarding components that use an SDK key to use the new one.
+	//  1. we should start a new SDK client*, but only for the anchor (design §4.1: one upstream
+	//     connection per env, on the anchor). Non-anchor server keys get envStreams + handler bundles
+	//     above, but no upstream client — matching today's mobile-key behavior.
+	//  2. we should tell all event forwarding components that use an SDK key to use the new one,
+	//     again only when it is the anchor (events collapse to the anchor per kind — design §4.3).
```
aaron-zeisler: Remove references to the design document from comments throughout this pull request.

internal/relayenv/env_context_impl.go
```diff
@@ -447,14 +450,16 @@ func (c *envContextImpl) addCredential(newCredential credential.SDKCredential) {
...
 	case config.SDKKey:
-		if !c.offline {
-			go c.startSDKClient(key, nil, false)
-		}
-		if c.metricsEventPub != nil { // metrics event publisher always uses SDK key
-			c.metricsEventPub.ReplaceCredential(key)
-		}
-		if c.eventDispatcher != nil {
-			c.eventDispatcher.ReplaceCredential(key)
+		if key == c.keyRotator.SDKKey() {
```
aaron-zeisler, line 453: Does keyRotator have an Anchor() function? I wonder if that would be more readable here than "SDKKey".

aaron-zeisler:
This is no longer necessary. It's been included in #712

PR #716, opened 2026-06-22
fix: Close orphaned and stale SDK clients during key rotation

keelerm84:
## Summary

SDK key rotation in the Relay Proxy could leak credentials and SDK clients in three scenarios surfaced while building out concurrent-keys support:

- **Re-anchoring expired the active primary.** Rotating an SDK key back into the primary slot while it was still within its deprecation grace period left it in the rotator's deprecated set, so the cleanup ticker would later expire the now-active primary. `swapPrimaryKey` now removes a re-anchored key from the deprecated set.
- **Already-expired grace never revoked the previous primary.** A rotation arriving with a grace period that had already elapsed neither added nor revoked the previous primary, so it kept authenticating indefinitely. The already-expired-grace branch of `updateSDKKey` now revokes the previous primary.
- **Orphaned and stale SDK clients.** `startSDKClient` builds the client before taking the env lock and then installed it unconditionally. A key revoked (or the environment closed) while the client was still being constructed left a live client -- open upstream connection and goroutines -- that nothing closed until shutdown; re-anchoring a key whose previous client was still alive leaked that stale client too. `startSDKClient` now closes the stale same-key client before replacing it on re-anchor, and closes (rather than installs) a freshly-built client when, by the time it holds the lock, the key has been revoked or the environment has been closed (checked against the rotator's tracked credentials).

Regression tests cover each path, including a gated-factory test that deterministically reproduces the install-after-revoke race. Validated under the race detector.

---

> [!NOTE]
> **Medium Risk**
> Touches credential rotation and SDK client lifecycle on a timing-sensitive path; mistakes could drop valid clients or leave connections open, but changes are narrowly scoped with race-detector-style tests.
> 
> **Overview**
> Fixes **SDK key rotation** leaks and incorrect credential lifetime behavior in the Relay Proxy.
> 
> **Credential rotator (regression tests)** documents and locks in two SDK-key fixes: rotations with an **already-expired grace** must **revoke the outgoing primary immediately**, and **re-anchoring** a key back to primary must **remove it from the deprecated set** so the cleanup ticker cannot expire the active anchor later.
> 
> **`startSDKClient`** no longer installs every successfully built client unconditionally. After the factory returns, it **drops and closes** clients when the environment is **closed** or the SDK key is **no longer tracked** (via new `sdkKeyIsActive`, with an exception for **undefined** keys so keyless envs still work). On install, it **closes any existing client** for the same key before replacing it (stale client on re-anchor). Successful-but-discarded inits still signal **ready** and log an info message instead of treating it as a hard failure.
> 
> **Tests** add rotator cases plus env tests for re-anchor stale close, a **gated factory** race (revoke while client is starting), and env-without-SDK-key behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3f8243e0e18be36521000da87847a4653b2f0998. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/relayenv/env_context_impl.go
+++ b/internal/relayenv/env_context_impl.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"slices"
 	"sync"
 	"time"
 
@@ -576,12 +577,7 @@ func (c *envContextImpl) startSDKClient(sdkKey config.SDKKey, readyCh chan<- Env
 // to avoid installing (and thereby leaking) a client for a key that was revoked while the client was
 // being constructed.
 func (c *envContextImpl) sdkKeyIsActive(sdkKey config.SDKKey) bool {
-	for _, cred := range c.keyRotator.AllCredentials() {
-		if cred == sdkKey {
-			return true
-		}
-	}
-	return false
+	return slices.Contains(c.keyRotator.AllCredentials(), credential.SDKCredential(sdkKey))
 }
 
 func (c *envContextImpl) GetPayloadFilter() config.FilterKey {
--- a/internal/relayenv/env_context_impl.go
+++ b/internal/relayenv/env_context_impl.go
@@ -500,7 +500,29 @@ func (c *envContextImpl) startSDKClient(sdkKey config.SDKKey, readyCh chan<- Env
 	client, err := c.sdkClientFactory(sdkKey, c.sdkConfig, c.sdkInitTimeout)
 	c.mu.Lock()
 	name := c.identifiers.GetDisplayName()
+	droppedInactive := false
+	if client != nil && (c.closed || (sdkKey.Defined() && !c.sdkKeyIsActive(sdkKey))) {
+		// startSDKClient builds the client before taking c.mu, so by the time we hold the lock the key
+		// may already have been revoked (rotated away) or the environment may have been closed. In
+		// either case the freshly-built client must be closed here rather than installed, otherwise its
+		// upstream connection and goroutines leak until env.Close() (and, once closed, nothing ever
+		// closes it). removeCredential cannot close it because the client was never in c.clients.
+		//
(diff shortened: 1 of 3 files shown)


internal/relayenv/env_context_impl_test.go
```diff
@@ -300,6 +301,174 @@ func TestChangeSDKKey(t *testing.T) {
...
+	assert.Equal(t, env.GetClient(), clientA1)
+
+	start := time.Unix(1000, 0)
+
+	// Rotate keyA -> keyB, deprecating keyA with an hour-long grace. keyA's client (clientA1) stays
+	// alive because keyA is still accepted during the grace window.
+	env.UpdateCredential(
+		NewCredentialUpdate(keyB).
+			WithTime(start).
+			WithGracePeriod(keyA, start.Add(1*time.Hour)))
```
aaron-zeisler, line 539: FYI: There will be a merge conflict in this code. #712 replaces `UpdateCredential()` with a new function `ReconcileCredentials()`.

internal/relayenv/env_context_impl.go
```diff
@@ -534,14 +557,33 @@ func (c *envContextImpl) startSDKClient(sdkKey config.SDKKey, readyCh chan<- Env
...
+// to avoid installing (and thereby leaking) a client for a key that was revoked while the client was
+// being constructed.
+func (c *envContextImpl) sdkKeyIsActive(sdkKey config.SDKKey) bool {
+	for _, cred := range c.keyRotator.AllCredentials() {
+		if cred == sdkKey {
+			return true
+		}
+	}
+	return false
+}
```
aaron-zeisler, line 581: Can this function (or the function's body) be replaced by slices.Contains()?

aaron-zeisler (approved):

PR #719, opened 2026-06-24
feat(credentials): wire RAC and offline handlers to ReconcileCredentials

aaron-zeisler:
## Summary

Completes Milestone 2 (anchor + multi-key auth). Wires both trusted-source handlers — the RAC stream (`relay/autoconfig_actions.go`) and the offline archive (`relay/filedata_actions.go`) — to the `ReconcileCredentials(AcceptedSet)` API, replacing the legacy single-key `UpdateCredential` path.

- Build the `AcceptedSet` from the full parsed key arrays (`params.AcceptedSDKKeys` / `AcceptedMobileKeys`, incl. per-key expiry) via the T3.b helper `BuildAcceptedSet`, then call `EnvContext.ReconcileCredentials`.
- Remove `CredentialUpdate` / `UpdateCredential` entirely from `EnvContext`; add `GetSDKKey()` returning the anchor.
- Surface deprecated/expiring SDK keys to the status endpoint via `Rotator.DeprecatedCredentials()` and `EnvContext.GetDeprecatedCredentials()`.
- Implement the design §9 malformed-credential policy: preserve previous credentials + reconnect the RAC stream (the offline path logs and preserves, with no reconnect).

Jira: [SDK-2547](https://launchdarkly.atlassian.net/browse/SDK-2547) · Epic: [SDK-2453](https://launchdarkly.atlassian.net/browse/SDK-2453)

## Background

The RAC and offline handlers previously called `UpdateCredential(update.WithGracePeriod(...))`, which only models a single expiring key. This wires them to `ReconcileCredentials`, which carries the full accepted-key array.

Semantic shifts in the new model:

1. **`GetCredentials()` is the full accepted set** — primary + all expiring keys (they still authenticate until expiry).
2. **`GetSDKKey()` returns the anchor** — `GetCredentials()` can hold several SDK keys, so the status endpoint uses `GetSDKKey()` for its `sdkKey` field to stay deterministic.
3. **`DeprecatedCredentials()` reports keys being phased out** — non-anchor accepted SDK keys carrying a future expiry (plus the legacy grace-period bucket during the transition). Expiring *mobile* keys are reported as accepted, not deprecated (no `expiringMobileKey` status field).

## Malformed-credential handling (design §9)

A structurally malformed credential payload — undefined anchor, an anchor absent from `sdkKeys[]`, or an array entry with an empty value — must **preserve the previous accepted set** and **reconnect the RAC stream** so the backend pushes a fresh `put` (RAC is one-way push with no NAK channel).

- **Validation lives in `BuildAcceptedSet`** (the single home for the anchor invariant) and runs at the **stream parse boundary** in `StreamManager`, before `MessageReceiver.Upsert` records the version. Validating before the version is recorded is essential: otherwise the reconnect's fresh `put` (same version) would be deduplicated away. Because the check sits ahead of dispatch, `put` and `patch` behave uniformly and `UpdateEnvironment` keeps its original void signature.
- **Cache safety:** a `put` carrying a malformed env does not corrupt the persistent auto-config cache. Valid envs and filters are persisted, each malformed env keeps its previously-cached entry (read-modify-write rather than a full `SetAll` replace), and an all-malformed `put` never wipes the cache.
- **Offline path** has no live stream, so it logs and preserves previous state without reconnecting.

## Notes for reviewers

- Merged the latest `feat/concurrent-keys`, which includes #716 ("Close orphaned and stale SDK clients during key rotation"). #716's `TestReAnchoringToKeyStillInGraceClosesStaleClient` encoded the old `UpdateCredential` client lifecycle; under `ReconcileCredentials` that leak is structurally impossible (re-anchoring to a still-accepted key reuses its client rather than spawning a new one), so the test is renamed to `...ReusesItsClient` and asserts the reuse-and-no-leak guarantee. The interaction with #716's re-queue work is captured on [SDK-2542](https://launchdarkly.atlassian.net/browse/SDK-2542).

## Follow-ups (blocked on this PR)

- [SDK-2603](https://launchdarkly.atlassian.net/browse/SDK-2603) — remove the now-unused legacy single-key rotation path from `Rotator` (the `TEMPORARY` lines in `DeprecatedCredentials` point here).
- [SDK-2604](https://launchdarkly.atlassian.net/browse/SDK-2604) — rename the anchor concept (`sdkKey` → `anchor`) across `Rotator`/`EnvContext`.
- [SDK-2609](https://launchdarkly.atlassian.net/browse/SDK-2609) - Integration scenarios for multi-key auth & rotation

[SDK-2547]: https://launchdarkly.atlassian.net/browse/SDK-2547?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **High Risk**
> Changes authentication credential ingestion, stream reconnect behavior, and persistent auto-config cache semantics—core relay security and availability paths.
> 
> **Overview**
> Wires **auto-config (RAC)** and **offline archive** handlers to **`ReconcileCredentials`** via **`BuildAcceptedSet`**, replacing the legacy **`UpdateCredential`** / grace-period path. **`EnvContext`** drops **`CredentialUpdate`** and adds **`GetSDKKey()`** / **`GetMobileKey()`** so the status endpoint can show a single anchor and primary mobile key deterministically.
> 
> **`Rotator.DeprecatedCredentials()`** now surfaces non-anchor SDK keys with future expiry from the reconcile model (for **`expiringSdkKey`** on status). **`GetCredentials()`** is treated as the full accepted set in tests and helpers (subset checks where expiring keys may also be present).
> 
> **Design §9** is implemented at the **`StreamManager`** boundary: **`validateCredentialPayload`** runs **`BuildAcceptedSet`** before **`Upsert`**, so bad payloads do not advance version or dispatch updates; the stream **restarts** to force a fresh put. **`handlePut`** still applies well-formed envs; **`persistPut`** read-modify-writes the cache so malformed envs keep last-good entries and all-malformed puts do not wipe **`SetAll`**.
> 
> **`BuildAcceptedSet`** is the single place for structural rules: undefined anchor, anchor missing from **`sdkKeys[]`**, and empty array values → **`MalformedCredentialSetError`**. **`ToParams`** no longer synthesizes an empty mobile key for server-only envs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a2d053cff1e42d53f97124c4efb9d517cef0b54f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/autoconfig/stream_manager_errors_test.go
+++ b/internal/autoconfig/stream_manager_errors_test.go
@@ -93,6 +93,13 @@ func TestMalformedCredentialPayloadPreservesEnvironmentCache(t *testing.T) {
 		env.SDKKey = envfactory.SDKKeyRep{Value: config.SDKKey("")} // undefined anchor
 		return env
 	}
+	// updated gives env a distinct SDK key value and a bumped version, so the test can tell a freshly
+	// persisted update apart from the seeded value.
+	updated := func(env envfactory.EnvironmentRep, newKey config.SDKKey) envfactory.EnvironmentRep {
+		env.SDKKey = envfactory.SDKKeyRep{Value: newKey}
+		env.Version++
+		return env
+	}
 	seedBothEnvs := func(t *testing.T, p streamManagerTestParams, cache *recordingCache) {
 		p.stream.Enqueue(makeEnvPutEvent(testEnv1, testEnv2))
 		require.Eventually(t, func() bool {
@@ -101,19 +108,26 @@ func TestMalformedCredentialPayloadPreservesEnvironmentCache(t *testing.T) {
 		}, time.Second, 10*time.Millisecond, "both envs should be cached after the clean put")
 	}
 
-	t.Run("mixed put keeps the valid and the malformed env in the cache", func(t *testing.T) {
+	t.Run("valid env is updated, malformed env keeps its previous value", func(t *testing.T) {
 		cache := &recordingCache{}
 		streamManagerTestWithCache(t, nil, cache, func(p streamManagerTestParams) {
 			p.startStream()
 			<-p.requestsCh
 			seedBothEnvs(t, p, cache)
 
-			p.stream.Enqueue(makeEnvPutEvent(testEnv1, malformed(testEnv2)))
+			// env1 carries a valid update to a new SDK key; env2 is malformed in the same put.
+			p.stream.Enqueue(makeEnvPutEvent(updated(testEnv1, "sdkkey1-rotated"), malformed(testEnv2)))
 			_ = helpers.RequireValue(t, p.requestsCh, time.Second, "timed out waiting for stream restart")
 
-			ids := cache.cachedEnvIDs()
-			assert.True(t, ids[testEnv1.EnvID], "the valid env must remain cached")
-			assert.True(t, ids[testEnv2.EnvID], "the malformed env must keep its previously-cached entry, not be dropped")
+			env1, ok1 := cache.cachedEnv(testEnv1.EnvID)
+			require.True(t, ok1, "the valid env must remain cached")
+			assert.Equal(t, config.SDKKey("sdkkey1-rotated"), env1.SDKKey.Value,
+				"the valid env's update must be persisted")
+
+			env2, ok2 := cache.cachedEnv(testEnv2.EnvID)
+			require.True(t, ok2, "the malformed env must keep its previously-cached entry, not be dropped")
+			assert.Equal(t, testEnv2.SDKKey.Value, env2.SDKKey.Value,
+				"the malformed env must retain its previous (valid) cached value, not the malformed one")
 		})
 	})
 
@@ -127,8 +141,11 @@ func TestMalformedCredentialPayloadPreservesEnvironmentCache(t *testing.T) {
 			p.stream.Enqueue(makeEnvPutEvent(malformed(testEnv1), malformed(testEnv2)))
 			_ = helpers.RequireValue(t, p.requestsCh, time.Second, "timed out waiting for stream restart")
 
-			ids := cache.cachedEnvIDs()
-			assert.True(t, ids[testEnv1.EnvID] && ids[testEnv2.EnvID], "an all-malformed put must not wipe the cache")
+			env1, ok1 := cache.cachedEnv(testEnv1.EnvID)
+			env2, ok2 := cache.cachedEnv(testEnv2.EnvID)
+			require.True(t, ok1 && ok2, "an all-malformed put must not wipe the cache")
+			assert.Equal(t, testEnv1.SDKKey.Value, env1.SDKKey.Value, "env1 keeps its previous value")
+			assert.Equal(t, testEnv2.SDKKey.Value, env2.SDKKey.Value, "env2 keeps its previous value")
 		})
 	})
 }
--- a/internal/autoconfig/stream_manager_errors_test.go
+++ b/internal/autoconfig/stream_manager_errors_test.go
@@ -85,33 +85,28 @@ func TestMalformedCredentialPayloadCausesStreamRestart(t *testing.T) {
 	})
 }
 
-// A put carrying a malformed credential payload must not be written to the persistent cache: doing so
-// would re-surface the rejected environment (skipped again) on the next cache load. The malformed env
-// is skipped in memory and the stream reconnects for a fresh put, which is what refreshes the cache.
-func TestMalformedCredentialPayloadInPutIsNotCached(t *testing.T) {
-	malformedEnv := testEnv1
+// A put that carries a malformed credential payload must still cache its valid environments, dropping
+// only the malformed one. Caching the rejected env would re-surface it (skipped again) on the next
+// cache load; failing to cache the valid envs would roll their updates back on a restart. The malformed
+// env is fetched fresh via the reconnect the put triggers.
+func TestMalformedCredentialPayloadIsExcludedFromCache(t *testing.T) {
+	validEnv := testEnv1
+	malformedEnv := testEnv2
 	malformedEnv.SDKKey = envfactory.SDKKeyRep{Value: config.SDKKey("")} // undefined anchor
 
-	t.Run("malformed put is not persisted", func(t *testing.T) {
-		cache := &recordingCache{}
-		streamManagerTestWithCache(t, nil, cache, func(p streamManagerTestParams) {
-			p.startStream()
-			<-p.requestsCh
-			p.stream.Enqueue(makeEnvPutEvent(malformedEnv))
-			_ = helpers.RequireValue(t, p.requestsCh, time.Second, "timed out waiting for stream restart")
-			assert.Equal(t, 0, cache.setAllCount(), "a malformed put must not be written to the cache")
-		})
-	})
+	cache := &recordingCache{}
+	streamManagerTestWithCache(t, nil, cache, func(p streamManagerTestParams) {
+		p.startStream()
+		<-p.requestsCh
+		p.stream.Enqueue(makeEnvPutEvent(validEnv, malformedEnv))
+		// The malformed env triggers a reconnect; observing it means handlePut has finished, including
+		// the cache write.
+		_ = helpers.RequireValue(t, p.requestsCh, time.Second, "timed out waiting for stream restart")
 
-	t.Run("valid put is persisted", func(t *testing.T) {
-		cache := &recordingCache{}
-		streamManagerTestWithCache(t, nil, cache, func(p streamManagerTestParams) {
-			p.startStream()
-			<-p.requestsCh
-			p.stream.Enqueue(makeEnvPutEvent(testEnv1))
-			assert.Eventually(t, func() bool { return cache.setAllCount() == 1 }, time.Second, 10*time.Millisecond,
-				"a valid put must be written to the cache")
-		})
+		require.GreaterOrEqual(t, cache.setAllCount(), 1, "the valid env in the put must still be cached")
+		cached := cache.lastCachedEnvIDs()
+		assert.Contains(t, cached, validEnv.EnvID, "the valid env must be cached")
+		assert.NotContains(t, cached, malformedEnv.EnvID, "the malformed env must be excluded from the cache")
 	})
 }
 
--- a/internal/autoconfig/stream_manager_errors_test.go
+++ b/internal/autoconfig/stream_manager_errors_test.go
@@ -148,6 +148,24 @@ func TestMalformedCredentialPayloadPreservesEnvironmentCache(t *testing.T) {
 			assert.Equal(t, testEnv2.SDKKey.Value, env2.SDKKey.Value, "env2 keeps its previous value")
 		})
 	})
+
+	t.Run("mixed put on an empty cache still persists the valid env", func(t *testing.T) {
+		cache := &recordingCache{}
+		streamManagerTestWithCache(t, nil, cache, func(p streamManagerTestParams) {
+			p.startStream()
+			<-p.requestsCh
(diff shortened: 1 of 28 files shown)


internal/credential/rotator.go
```diff
@@ -158,6 +158,25 @@ func (r *Rotator) DeprecatedCredentials() []SDKCredential {
...
+	defer r.mu.RUnlock()
+	out := r.deprecatedCredentials()
+	for key, info := range r.acceptedSDKKeys {
+		if info.expiry != nil && key != r.primarySdkKey {
+			out = append(out, key)
+		}
+	}
+	return out
+}
+
```
aaron-zeisler, line 179: Why exactly is this necessary? The comment says "without caring which rotation path produced it" - Is deprecatedCredentials being populated differently by two different code paths? Is this referring to the legacy path vs the new concurrent-keys path? Or some other difference?

internal/relayenv/env_context.go
```diff
@@ -77,13 +40,10 @@ type EnvContext interface {
...
-	// with a grace period.
-	//
-	// This is the legacy single-credential rotation API. It is retained while the action handlers still
-	// drive rotation through it; the full-set ReconcileCredentials below will take over once the handlers
-	// are migrated to it.
-	UpdateCredential(update *CredentialUpdate)
+	// GetSDKKey returns the anchor SDK key — the primary key that owns the upstream connection.
+	// Use this when you need exactly the anchor (e.g. the status endpoint's sdkKey field) rather
+	// than the full accepted set returned by GetCredentials.
+	GetSDKKey() config.SDKKey
```
aaron-zeisler, line 46: This should be called "GetAnchorKey". I think the "SDKKey" function on the rotator should also be called "AnchorKey" or "GetAnchorKey".

internal/credential/rotator.go
```diff
@@ -151,11 +151,30 @@ func (r *Rotator) deprecatedCredentials() []SDKCredential {
...
+	// the legacy path is removed (SDK-2603) these buckets are always empty; delete this line and the
+	// deprecatedCredentials helper, leaving only the accepted-with-expiry logic below.
+	out := r.deprecatedCredentials()
+
+	for key, info := range r.acceptedSDKKeys {
+		if info.expiry != nil && key != r.primarySdkKey {
+			out = append(out, key)
+		}
+	}
+	return out
```
aaron-zeisler, line 177: I'll clean this up when I remove all the old (single-key) code in SDK-2603

internal/relayenv/env_context.go
aaron-zeisler: I'll rename "sdk key" to "anchor" throughout the code (where it makes sense) in SDK-2604

internal/projmanager/project_router.go
```diff
@@ -53,16 +53,16 @@ func (e *ProjectRouter) AddEnvironment(params envfactory.EnvironmentParams) {
...
 func (e *ProjectRouter) UpdateEnvironment(params envfactory.EnvironmentParams) {
 	proj := params.Identifiers.ProjKey
 	manager, ok := e.managers[proj]
 	if ok {
 		manager.UpdateEnvironment(params)
-	} else {
-		e.loggers.Errorf("precondition violation: received updated config for (%s), but environment was never added", params.Identifiers.GetDisplayName())
+		return
 	}
+	e.loggers.Errorf("precondition violation: received updated config for (%s), but environment was never added", params.Identifiers.GetDisplayName())
```
aaron-zeisler: This change looks unrelated to the rest of the changes in the pull request. Although I understand the "else" block here can be cleanup up, I'd rather not change code if it has no impact on the functionality of our specific goal.

internal/credential/accepted_set.go
```diff
@@ -54,25 +54,44 @@ func (s AcceptedSet) hasMobileKey(key config.MobileKey) bool {
...
+// value field is empty. kind is "sdkKeys" or "mobileKeys"; identifier is the key's identifier
+// string (may be empty for old-format payloads that synthesize from the singular fields).
+func NewEmptyCredentialError(kind, identifier string) *MalformedCredentialSetError {
+	if identifier == "" {
+		return &MalformedCredentialSetError{
+			msg: fmt.Sprintf("malformed credential set: %s entry has an empty value", kind),
+		}
+	}
+	return &MalformedCredentialSetError{
+		msg: fmt.Sprintf("malformed credential set: %s entry %q has an empty value", kind, identifier),
```
aaron-zeisler, line 102: I see similar string constants were added in `autoconfig/errors_and_messages.go` and `relay/autoconfig_actions.go`. Are they all really different, or can we define these messages in just one spot?

internal/autoconfig/stream_manager_errors_test.go
```diff
@@ -30,6 +33,106 @@ func eventShouldCauseStreamRestart(t *testing.T, event httphelpers.SSEEvent) {
...
+	}
+	seedBothEnvs := func(t *testing.T, p streamManagerTestParams, cache *recordingCache) {
+		p.stream.Enqueue(makeEnvPutEvent(testEnv1, testEnv2))
+		require.Eventually(t, func() bool {
+			ids := cache.cachedEnvIDs()
+			return ids[testEnv1.EnvID] && ids[testEnv2.EnvID]
+		}, time.Second, 10*time.Millisecond, "both envs should be cached after the clean put")
+	}
+
+	t.Run("mixed put keeps the valid and the malformed env in the cache", func(t *testing.T) {
```
keelerm84: Are these tests actually valid? Since it seeds with the same environments it is trying to replace, I'm not sure it is correctly asserting success or failure.

internal/autoconfig/stream_manager_errors_test.go
```diff
@@ -30,6 +33,106 @@ func eventShouldCauseStreamRestart(t *testing.T, event httphelpers.SSEEvent) {
...
+			p.stream.Enqueue(makeEnvPutEvent(testEnv1, malformed(testEnv2)))
+			_ = helpers.RequireValue(t, p.requestsCh, time.Second, "timed out waiting for stream restart")
+
+			ids := cache.cachedEnvIDs()
+			assert.True(t, ids[testEnv1.EnvID], "the valid env must remain cached")
+			assert.True(t, ids[testEnv2.EnvID], "the malformed env must keep its previously-cached entry, not be dropped")
+		})
+	})
+
+	t.Run("all-malformed put does not wipe the cache", func(t *testing.T) {
```
keelerm84, line 134: Same with this test.

internal/credential/accepted_set.go
aaron-zeisler: These messages are at different application layers - I'll keep them separate.

keelerm84: Personally, I'm not a fan of constants that only occur in a single place. So I think a lot of those examples don't need to exist.

I think just inlining the messages like this is fine IMO.

internal/credential/rotator.go
```diff
@@ -151,11 +151,30 @@ func (r *Rotator) deprecatedCredentials() []SDKCredential {
 	return deprecated
 }
 
-// DeprecatedCredentials returns deprecated credentials (not expired or primary.)
+// DeprecatedCredentials returns the SDK keys being phased out: any non-anchor accepted SDK key that
+// carries a future expiry (the reconcile model stores per-key expiry as data on the accepted entry).
+// EnvContext.GetDeprecatedCredentials delegates here so the status endpoint can surface its
+// expiringSdkKey field.
+//
+// Reconcile-path mobile keys with an expiry are intentionally excluded: there is no expiringMobileKey
```
keelerm84: This comment is kind of confusing. Also why is there no expiry on mobile keys? Is that coming in a subsequent PR?

internal/envfactory/reconcile_helper_test.go
```diff
@@ -164,16 +168,49 @@ func TestBuildAcceptedSet_AnchorNotInArray(t *testing.T) {
...
+	require.Error(t, err)
+	var malformed *credential.MalformedCredentialSetError
+	require.True(t, errors.As(err, &malformed))
+	assert.Contains(t, malformed.Error(), "not present in sdkKeys[]")
+}
+
+// TestBuildAcceptedSet_AnchorNotInNewFormatArray exercises the realistic wire path: a new-format
+// EnvironmentRep (sdkKeys[] present) whose sdkKey.value is absent from the array must be rejected after
+// ToParams, not silently synthesized in.
+func TestBuildAcceptedSet_AnchorNotInNewFormatArray(t *testing.T) {
```
keelerm84: Is this not redundant with TestBuildAcceptedSet_AnchorNotInArray?

relay/endpoints_status.go
```diff
@@ -46,17 +46,19 @@ func statusHandler(relay *Relay) http.Handler {
...
+			// (primary + expiring), so iterating it would give a non-deterministic result.
+			if key := clientCtx.GetSDKKey(); key.Defined() {
+				status.SDKKey = sdks.ObscureKey(string(key))
+			}
 			for _, c := range clientCtx.GetCredentials() {
 				switch c := c.(type) {
-				case config.SDKKey:
-					status.SDKKey = sdks.ObscureKey(string(c))
 				case config.MobileKey:
 					status.MobileKey = sdks.ObscureKey(string(c))
```
keelerm84: **[review] `mobileKey` is still non-deterministic with multiple accepted mobile keys.**

This PR fixed exactly this non-determinism for the SDK key by adding `GetSDKKey()` just above (L51-53), but the mobile case still iterates `GetCredentials()` and keeps the last `config.MobileKey` seen. With more than one accepted mobile key, the reported `mobileKey` depends on map iteration order. The Rotator already exposes `MobileKey()` (the primary) -- mirror the SDK fix by adding `GetMobileKey()` to the `EnvContext` interface and using it here. Cosmetic (status output only; routing unaffected).

aaron-zeisler: Thanks. I _just_ fixed this in response to Bugbot's latest issue.

relay/endpoints_status.go
```diff
@@ -46,17 +46,20 @@ func statusHandler(relay *Relay) http.Handler {
...
 			for _, c := range clientCtx.GetCredentials() {
-				switch c := c.(type) {
-				case config.SDKKey:
-					status.SDKKey = sdks.ObscureKey(string(c))
-				case config.MobileKey:
-					status.MobileKey = sdks.ObscureKey(string(c))
-				case config.EnvironmentID:
-					status.EnvID = string(c)
+				if envID, ok := c.(config.EnvironmentID); ok {
+					status.EnvID = string(envID)
```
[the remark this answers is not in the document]
aaron-zeisler: I'm deferring this work to SDK-2534 - In that task, I'll update all the status endpoint code to support multiple keys.

internal/envfactory/reconcile_helper_test.go
aaron-zeisler: Thanks, yes, I've removed this test (and another that was redundant)

internal/credential/rotator.go
aaron-zeisler: Mobile keys do have an expiration, and that expiration is used when managing the keys and when determining how to handle an incoming request from an SDK. This specific function "DeprecatedCredentials()" is used by the status endpoints, and those endpoints haven't yet been updated to work with concurrent keys. I'm planning to perform work on the status endpoints in SDK-2534 - that's when the expiration date/time for mobile keys will be exposed via this function.

internal/autoconfig/stream_manager_errors_test.go
aaron-zeisler: Thanks, I've updated both the tests.

keelerm84 (approved):

PR #720, opened 2026-06-25
feat(relayenv): re-anchor mechanism

aaron-zeisler:
## Summary

Implements **T2.c** — the re-anchor mechanism (SDK-2542) for concurrent SDK keys. When `sdkKey.value` changes, relay now swaps its single upstream SDK client to the new anchor **synchronously**, preserving downstream connections and rolling back cleanly on failure.

Jira: [SDK-2542](https://launchdarkly.atlassian.net/browse/SDK-2542) · Epic: [SDK-2453](https://launchdarkly.atlassian.net/browse/SDK-2453) · Milestone M3.

Targets `feat/concurrent-keys` (not v8). Builds on the T0 PoC findings and the store-handover spike.

## Background

The design (`.agent-docs/concurrent-keys/phase1-design.md` §7) and the T0 PoC established that re-anchoring is **not** a transparent side-effect of today's code: flipping the anchor pointer before the new client initializes leaves `GetClient()` nil mid-swap (PoC H6) and strands the env on init failure (H7); the in-memory store is rebuilt empty when the new client constructs (H1/H5). This PR closes those gaps.

## Changes (one commit per logical step)

1. **`test(store)`** — bring over the three store-handover real-client spike tests (one rewritten as `TestRealClient_HandoverPreservesUnderlyingStore`).
2. **`feat(store)`** — `SSERelayDataStoreAdapter.Build` reuses its existing wrapper on a second client construction (store handover → no empty-store window). The wrapper is **refcounted**: only the final `Close` tears down the underlying store, so the retiring client's `Close` can't pull the store out from under the new anchor (design §7 lifecycle caveat).
3. **`feat(credential)`** — `Reconcile` no longer flips the anchor; it returns a `ReconcileResult` (`AnchorChange` + `MobilePrimaryRepoint`), and a new `CommitAnchor` moves the pointer. The new anchor is stripped from `additions` in Case A so `addCredential`'s async `startSDKClient` can't race the synchronous build.
4. **`feat(relayenv)`** — synchronous re-anchor with **Case A** (build new client + store handover + wait `Initialized` → commit → `ReplaceCredential` → big-segment stub; rollback on init failure) and **Case B** (reuse the existing client → commit → `ReplaceCredential`). Order: add → re-anchor → remove, serialized by a dedicated `reconcileMu`.
5. **`test(relayenv)`** — integration tests: Case A success, Case A init-failure rollback, Case B reuse (no new client), mobile-primary repoint.
6. **`docs`** — document Case B in design §7.

## Acceptance criteria coverage

| Criterion | Where |
|---|---|
| Case A success (build, init, flip, store handover, old client retained) | `TestReanchorSync_CaseA_BuildsNewClientAndMovesAnchor` |
| Case A init failure → rollback (anchor stays, old client serves, structured error) | `TestReanchorSync_CaseA_InitFailureRollsBack` |
| Case B reuse (no new Build, anchor flips) | `TestReanchorSync_CaseB_ReusesExistingClient` |
| Store handover (Build reuse; Close doesn't tear down shared store) | `TestRealClient_HandoverPreservesUnderlyingStore`, `TestReanchorPoC_H5_StoreSurvivesReAnchor`, H1 sub-test |
| Mobile primary switch repoints event dispatcher; non-primary does not | `TestReanchorSync_MobilePrimaryRepoint_AlreadyAcceptedKey` + PR #712's gate |
| No orphan clients (Case A doesn't trip #716 guard; Case B spawns nothing) | Case A / Case B tests |

PoC tests for H1 and H5 were inverted from "asserts broken behavior" to "asserts the fix" and retained as regressions.

## Out of scope

- **T2.d (SDK-2543)** — big-segment sync re-wire. A clean call site (`reanchorBigSegmentSync`) is defined as a no-op (matches today's behavior, no regression); T2.d fills it in.
- **T2.e** — handler fan-out.

## Post-merge note (catch-up onto `feat/concurrent-keys`)

The base moved 11 commits ahead after this branch was cut, including the `sdkKey→anchor` rename (#722), the legacy single-key rotation path removal (#723), the `SDKKeyParams`/`MobileKeyParams` builder + `AcceptedKeys()` reshaping (#729/#730), and the RAC/offline wiring to `ReconcileCredentials` (#719). This branch now catch-up **merges** `feat/concurrent-keys` and re-applies the T2.c intent on top of the refactored API:

- `Reconcile` still defers the anchor flip to `CommitAnchor`; the in-place flip that the rebased `reconcileSDKKeys` introduced was removed. Case A still strips the new anchor from `additions`.
- The base unit tests that assumed Reconcile flips in place (`TestAcceptedKeys`, `TestReconcileAlreadyExpiredKeyIsIgnoredOnAdd`) were updated to the deferred-flip contract (`CommitAnchor` to observe the anchor).
- PoC **H6/H7** were inverted to assert the fix (no nil swap-window; rollback keeps the old anchor authoritative), completing the same treatment H1/H5 received.
- **#719 is now merged.** Its `ReconcileCredentials`-based regression `TestReAnchoringToKeyStillInGraceReusesItsClient` and this PR's `TestReanchorSync_CaseB_ReusesExistingClient` both cover Case B reuse; they are complementary (the former focuses on no-leak/close-of-revoked-key, the latter on no-new-build + anchor flip) and are both retained.

### Review fix — re-anchor rollback no longer poisons `initErr`

A review of the merge caught a regression in the rollback path: on a failed re-anchor to an **invalid** new key the SDK returns `ErrInitializationFailed`, and `reanchor` was storing that in `c.initErr`. The request middleware treats `GetInitError() == ErrInitializationFailed` as an env-level auth failure and returns **401 for every request** — which would have rejected all traffic to an environment that was still healthy on the previous anchor, defeating the rollback. Fixed: `initErr` is left untouched on rollback (the env is healthy on the old anchor); the failure still surfaces via the structured `Errorf` log. The Case A init-failure and H7 tests now assert `GetInitError()` stays nil and that the failure is logged.

### Review fix — serialize the credential cleanup ticker against re-anchor

A second review finding (MEDIUM): the cleanup ticker path (`cleanupExpiredCredentials` → `triggerCredentialChanges`) drained the rotator's `StepTime` queue and ran `addCredential`/`removeCredential` **without** holding `reconcileMu`, so a credential expiry firing during an in-flight synchronous re-anchor ran concurrently with the reanchor sequence. Both paths drain the same `StepTime` queue (the ticker could steal additions a reconcile just queued) and a ticker `removeCredential` could close a client mid-reanchor. The synchronous re-anchor widened this window versus the pre-T2.c base. Fixed: `triggerCredentialChanges` now takes `reconcileMu` for its whole `StepTime` + add/remove pass, serializing the ticker against `reconcileCredentials` the same way concurrent reconciles already are (no re-entrancy — `reconcileCredentials` never calls that path). New regression `TestReanchorSync_CredentialExpiryDuringReanchorIsSerialized` wedges a re-anchor open mid-build and asserts the ticker blocks until it releases.

### Review fix (LOW) — re-anchor peripheral install branches on one signal

`reanchor` now gates peripheral install (envStreams / handler / connection-mapper wiring) solely on `change.NewAnchorPreviouslyAccepted` rather than on the Case A/B `c.clients[newAnchor] != nil` branch. The two signals always agree today (a client can only exist for a previously-accepted key, since `removeCredential` deletes the client in lockstep with the rotator dropping the key), but branching peripheral setup on the rotator signal means a brand-new anchor is never stranded without routing even if that invariant were ever broken.

[SDK-2542]: https://launchdarkly.atlassian.net/browse/SDK-2542?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **High Risk**
> Changes core credential reconciliation, upstream SDK client lifecycle, and shared data-store refcounting during anchor rotation—failure or concurrency bugs could break serving or auth for whole environments.
> 
> **Overview**
> Implements **synchronous SDK anchor re-anchoring** when `sdkKey.value` changes: the upstream client swap no longer flips the anchor during `Reconcile` or races async `startSDKClient`.
> 
> **Credential rotator** now returns `ReconcileResult` (`AnchorChange`, `MobilePrimaryRepoint`) and defers the anchor move to **`CommitAnchor`**; brand-new anchors are stripped from `additions` so the env owns client build. **`RevertAnchorChange`** realigns the accepted set on Case A rollback.
> 
> **`reconcileCredentials`** runs add → synchronous **`reanchor`** → remove under **`reconcileMu`** (also held by the cleanup ticker’s `triggerCredentialChanges`). **`reanchor`** builds or reuses the new anchor client before commit, calls **`ReplaceCredential`** on events/metrics, rolls back without poisoning **`initErr`**, and handles mobile-primary repoint when the new primary was already accepted.
> 
> **Store handover**: `SSERelayDataStoreAdapter.Build` reuses the existing wrapper with **refcounted** `Close` so the retiring client does not tear down shared in-memory data.
> 
> Design §7 documents Case A vs Case B; PoC tests assert fixed behavior; new integration and real-client tests cover success, rollback, reuse, serialization, and offline commit-without-build. Big-segment re-wire remains a stub (T2.d).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2fc4f209682d9932405890dc4103119b25702f1b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/credential/rotator.go
+++ b/internal/credential/rotator.go
@@ -2,6 +2,7 @@ package credential
 
 import (
 	"maps"
+	"slices"
 	"sync"
 	"time"
 
@@ -236,9 +237,9 @@ type ReconcileResult struct {
 //     synchronous re-anchor must register the credential mappings (envStreams, handlers, connection
 //     mapping), construct and initialize a new SDK client, then invoke CommitAnchor + ReplaceCredential.
 //   - true (the anchor is a previously-accepted key): the new anchor was already accepted (typically
-//     a former anchor still in its grace period). Peripherals are already in place and a client may
-//     already exist; the synchronous re-anchor reuses it (or constructs one only if missing — see the
-//     re-anchor sequence in env_context_impl.go), then invokes CommitAnchor + ReplaceCredential.
+//     a former anchor still in its grace period). Its credential mappings are already registered and a
+//     client may already exist; the synchronous re-anchor reuses it (or constructs one only if missing
+//     — see the re-anchor sequence in env_context_impl.go), then invokes CommitAnchor + ReplaceCredential.
 type AnchorChange struct {
 	PreviousAnchor              config.SDKKey
 	NewAnchor                   config.SDKKey
@@ -288,7 +289,7 @@ func (r *Rotator) Reconcile(set AcceptedSet, now time.Time) ReconcileResult {
 		// addition normally, its async startSDKClient would race the synchronous build. When the anchor
 		// is a previously-accepted key it was already in acceptedSDKKeys, so reconcileAcceptedKeys did
 		// not add it — no strip needed.
-		r.additions = removeCredentialFromList(r.additions, newAnchor)
+		r.additions = slices.DeleteFunc(r.additions, func(c SDKCredential) bool { return c == newAnchor })
 	}
 
 	previousMobile := r.primaryMobileKey
@@ -308,15 +309,6 @@ func (r *Rotator) Reconcile(set AcceptedSet, now time.Time) ReconcileResult {
 	r.reconcileEnvironmentID(set)
 
 	return result
-}
-
-func removeCredentialFromList(list []SDKCredential, target SDKCredential) []SDKCredential {
-	for i, c := range list {
-		if c == target {
-			return append(list[:i], list[i+1:]...)
-		}
-	}
-	return list
 }
 
 // CommitAnchor atomically moves the rotator's SDK anchor pointer to the given key. The caller
--- a/internal/credential/rotator.go
+++ b/internal/credential/rotator.go
@@ -524,7 +524,6 @@ func (r *Rotator) CommitAnchor(key config.SDKKey) {
 	r.primarySdkKey = key
 }
 
-
 // reconcilableKey constrains the generic reconcile helper to a comparable credential (so it can key a
 // map) that is also an SDKCredential (so it can be logged and appended to the credential lists).
 type reconcilableKey interface {
--- a/internal/credential/rotator.go
+++ b/internal/credential/rotator.go
@@ -377,6 +377,7 @@ func reconcileAcceptedKeys[K reconcilableKey](
 	}
 }
 
+// reconcileSDKKeys diffs the desired SDK keys against the accepted set and applies the result via
 // reconcileAcceptedKeys. The set is trusted as well-formed: BuildAcceptedSet / the builder guarantee
 // the anchor is present and permanent (WithAnchor forces a nil expiry), so no special handling is
 // needed here. The caller must hold the write lock.
--- a/internal/credential/rotator.go
+++ b/internal/credential/rotator.go
@@ -399,6 +399,42 @@ func (r *Rotator) StepTime(now time.Time) (additions []SDKCredential, expiration
 	return additions, expirations
 }
 
+// ReconcileResult signals state changes that the caller must apply synchronously rather than rely
+// on the normal addCredential / removeCredential flow driven by StepTime.
+//
+// AnchorChange is non-nil when the SDK anchor changed during Reconcile. The rotator does NOT flip
+// its anchor pointer in that case — the caller must drive the synchronous re-anchor sequence
+// (build the new anchor's SDK client if one does not exist, wait for Initialized, then invoke
+// CommitAnchor to atomically move the pointer, then call ReplaceCredential on the event dispatcher
+// and metrics publisher, then re-wire big-segment sync). See .agent-docs/concurrent-keys/phase1-design.md §7.
+//
+// MobilePrimaryRepoint is non-nil when the primary mobile key changed AND the new primary was
+// already in the accepted set. In that case it does not appear in StepTime's additions list and
+// addCredential's primary-mobile gate will not fire for it, so the caller must invoke
+// eventDispatcher.ReplaceCredential synchronously. When nil, either the primary mobile key did not
+// change, or it changed to a newly-accepted key — in which case the normal addCredential path
+// handles the ReplaceCredential call via the existing gate.
+type ReconcileResult struct {
+	AnchorChange         *AnchorChange
+	MobilePrimaryRepoint *config.MobileKey
+}
+
+// AnchorChange describes an SDK anchor transition produced by Reconcile.
+//
+// NewAnchorPreviouslyAccepted distinguishes the two re-anchor paths described in design §7:
+//   - false (Case A): the new anchor was not previously in the accepted set. The synchronous
+//     re-anchor must perform peripheral setup (envStreams, handlers, connection mapping), construct
+//     and initialize a new SDK client, then invoke CommitAnchor + ReplaceCredential.
+//   - true (Case B): the new anchor was already accepted (typically a former anchor still in its
+//     grace period). Peripherals are already in place and a client may already exist; the
+//     synchronous re-anchor reuses it (or constructs one only if missing — see Case A vs B in
+//     env_context_impl.go), then invokes CommitAnchor + ReplaceCredential.
+type AnchorChange struct {
+	PreviousAnchor              config.SDKKey
+	NewAnchor                   config.SDKKey
+	NewAnchorPreviouslyAccepted bool
+}
+
 // Reconcile updates the rotator to match set. The set names its own anchor (the primary SDK key) and
 // primary mobile key. It diffs the desired accepted set against the current one and queues additions
 // and expirations (drained by the next StepTime call); keys newly present are accepted, and keys no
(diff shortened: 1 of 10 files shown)


internal/credential/rotator.go
```diff
@@ -209,6 +209,42 @@ func (r *Rotator) StepTime(now time.Time) (additions []SDKCredential, expiration
...
 }
 
+// ReconcileResult signals state changes that the caller must apply synchronously rather than rely
+// on the normal addCredential / removeCredential flow driven by StepTime.
+//
+// AnchorChange is non-nil when the SDK anchor changed during Reconcile. The rotator does NOT flip
+// its anchor pointer in that case — the caller must drive the synchronous re-anchor sequence
+// (build the new anchor's SDK client if one does not exist, wait for Initialized, then invoke
+// CommitAnchor to atomically move the pointer, then call ReplaceCredential on the event dispatcher
+// and metrics publisher, then re-wire big-segment sync). See .agent-docs/concurrent-keys/phase1-design.md §7.
```
aaron-zeisler: Remove all project-specific references to the documentation.

internal/credential/rotator.go
```diff
@@ -220,13 +256,82 @@ func (r *Rotator) StepTime(now time.Time) (additions []SDKCredential, expiration
...
+	r.reconcileSDKKeys(set, now)
+
+	if result.AnchorChange != nil && !result.AnchorChange.NewAnchorPreviouslyAccepted {
+		// Case A: reconcileAcceptedKeys just appended the brand-new anchor to r.additions. Strip
+		// it — the synchronous re-anchor sequence in env_context_impl owns the new anchor's setup
+		// (peripherals + client build + flip + ReplaceCredential). If addCredential drained this
+		// addition normally, its async startSDKClient would race the synchronous build. In Case B
+		// (anchor previously accepted) the anchor was already in acceptedSDKKeys so
+		// reconcileAcceptedKeys did not add it — no strip needed.
+		r.additions = removeCredentialFromList(r.additions, newAnchor)
```
aaron-zeisler: Is it possible to prevent reconcileSDKKeys from adding the anchor to additions? Would that make the logic more or less complex?

internal/relayenv/env_context_impl.go
```diff
@@ -123,6 +123,12 @@ type envContextImpl struct {
 	connectionMapper          ConnectionMapper
 	offline                   bool
 	closed                    bool
+
+	// reconcileMu serializes reconcileCredentials calls, including the synchronous re-anchor
+	// sequence inside them. Held separately from mu so that GetClient / GetStore / GetEvaluator /
+	// addCredential continue to run during the (potentially seconds-long) SDK client construction
+	// in a Case A re-anchor.
```
aaron-zeisler: * Does this comment mean to say that is "synchronizes" reconcileCredentials calls?
* This comment refers to "Case A", which is defined in the design document. Please refer to it more literally (ie "the anchor is a new key").

internal/credential/rotator.go
aaron-zeisler: Also remove "Case A" and "Case B" from comments throughout the PR, and replace those words with more explicit terms (ie "the anchor is a new key" and "the anchor is a previously-accepted key").

internal/relayenv/env_context_impl.go
```diff
@@ -605,16 +593,279 @@ func (c *envContextImpl) ReconcileCredentials(newSet credential.AcceptedSet) {
...
+	}
+
+	if offline {
+		// In offline mode there is no upstream client to build. Just commit + ReplaceCredential.
+		c.commitReanchor(newAnchor, previousAnchor, "offline — no client build")
+		return true
+	}
+
+	return c.buildAndCommitNewAnchorClient(newAnchor, previousAnchor)
 }
```
aaron-zeisler, line 747: This function contains a lot of mutex locking and unlocking, and it calls down into functions that perform more locking and unlocking. It looks like the commitReanchor() function is required (on top of commitReanchorLocked()) just because it must apply and release a mutex lock. Can we simply lock the mutex near the top of the reanchor() function and immediately mu.Unlock(), so the mutex is locked throughout all of the downstream functions?

The only part that I see that doesn't require the lock is lines 733-744, at the beginning of the buildAndCommitNewAnchorClient() function, when the client is built for the new anchor. It might be fine if the lock is held through that action, depending on the performance of c.sdkClientFactory().

internal/relayenv/env_context_impl.go
```diff
@@ -605,16 +593,263 @@ func (c *envContextImpl) ReconcileCredentials(newSet credential.AcceptedSet) {
...
+// committed, and the failure is logged. initErr is deliberately not set on failure: it feeds the
+// request middleware, so setting it to the new anchor's ErrInitializationFailed would 401 an env that
+// is serving fine on the previous anchor.
+//
+// Returns true if the anchor was committed, false if it rolled back (init failure or the env closed
+// mid-build).
+func (c *envContextImpl) buildAndCommitNewAnchorClient(newAnchor, previousAnchor config.SDKKey) bool {
+	c.mu.Unlock()
+	client, err := c.sdkClientFactory(newAnchor, c.sdkConfig, c.sdkInitTimeout)
+	c.mu.Lock()
```
aaron-zeisler: This quick unlock-and-lock is silly. Let's put lines 730 and 733-743 into a new function called buildNewAnchorClient() - that function shouldn't need any mutex lock, so it can be called in reanchor() before the mutex lock, just after line 685. That should further simplify the locking and unlocking. Then we can rename this function commitNewAnchorClient() and it can take the client as input.

aaron-zeisler:
This work has been completed by Matthew in a series of smaller, broken-down pull requests

PR #724, opened 2026-06-26
feat(status): add sdkKeys[] and mobileKeys[] arrays to status endpoint

aaron-zeisler:
> [!NOTE]
> **Closed without merging — superseded by a stack of smaller PRs.** This change grew to mix a credential-model refactor with the endpoint feature, so it was split into:
> - #728 — fix: reject primary mobile key absent from `mobileKeys[]` (independent correctness fix)
> - #729 → #730 → #731 — the feature stack (model: store identifiers → expose accepted set → surface on `/status`)
>
> Combined, those reconstruct the reviewed code here. This branch is kept as reference until the stack lands.

---

## Summary

Implements [SDK-2534](https://launchdarkly.atlassian.net/browse/SDK-2534) (T4 of concurrent-keys epic [SDK-2453](https://launchdarkly.atlassian.net/browse/SDK-2453)): adds `sdkKeys[]` and `mobileKeys[]` array fields to the `/status` endpoint response, surfacing the **full accepted key set** for each environment. The existing scalar `sdkKey`/`mobileKey` fields are preserved for back-compat and now designate which array entry is the anchor / primary.

Also makes the legacy `expiringSdkKey` field deterministic: the previous code overwrote on each loop iteration (arbitrary with multiple expiring keys); it now reports the key with the soonest expiry.

## Background

The concurrent-keys epic (SDK-2453) lets an environment accept multiple SDK keys and mobile keys at once. Per the backend tech spec and RAC payload spec, the `sdkKeys`/`mobileKeys` arrays are the authoritative **full accepted set, including the default/anchor key** (which is also duplicated on the legacy scalar fields). The status endpoint mirrors that: the arrays carry every accepted key; the scalars designate the anchor / primary.

The wire format carries a non-secret `key` identifier alongside each credential's `value`. This PR threads that identifier through to the status response. The identifier is genuinely optional — manual configuration and old-format (pre-concurrent-keys) payloads carry none — so it is modeled as a pointer and omitted from the JSON when absent.

## Response schema

`EnvironmentStatusRep` gains two arrays plus a `KeyStatus` element type. The scalar fields are unchanged.

```go
// KeyStatus is one accepted credential in the sdkKeys[] / mobileKeys[] arrays.
type KeyStatus struct {
    Key    string `json:"key,omitempty"`    // non-secret identifier; omitted when the source carried none
    Value  string `json:"value"`            // obscured credential secret
    Expiry *int64 `json:"expiry,omitempty"` // Unix-millis expiry; omitted for permanent keys
}

type EnvironmentStatusRep struct {
    SDKKey           string               `json:"sdkKey"`                   // obscured ANCHOR (designates which SDKKeys entry is the anchor)
    SDKKeys          []KeyStatus          `json:"sdkKeys"`                  // full server-key set, incl. anchor; always present, always ≥1
    EnvID            string               `json:"envId,omitempty"`
    EnvKey           string               `json:"envKey,omitempty"`
    EnvName          string               `json:"envName,omitempty"`
    ProjKey          string               `json:"projKey,omitempty"`
    ProjName         string               `json:"projName,omitempty"`
    MobileKey        string               `json:"mobileKey,omitempty"`      // obscured PRIMARY mobile key
    MobileKeys       []KeyStatus          `json:"mobileKeys"`               // full mobile-key set, incl. primary; always present, may be empty
    ExpiringSDKKey   string               `json:"expiringSdkKey,omitempty"` // soonest-expiring non-anchor SDK key (deterministic)
    Status           string               `json:"status"`
    ConnectionStatus ConnectionStatusRep  `json:"connectionStatus"`
    DataStoreStatus  DataStoreStatusRep   `json:"dataStoreStatus"`
    BigSegmentStatus *BigSegmentStatusRep `json:"bigSegmentStatus,omitempty"`
}
```

Array entry order is unspecified. The arrays are always present (never null): `sdkKeys` always contains at least the anchor; `mobileKeys` is empty for an environment with no mobile key (e.g. server-side only).

### Example (environment mid-rotation, an extra service key + an expiring key)

```json
"environment1": {
  "sdkKey": "sdk-********-****-****-****-*******00001",
  "sdkKeys": [
    { "key": "default",   "value": "sdk-********-****-****-****-*******00001" },
    { "key": "service-a", "value": "sdk-********-****-****-****-*******00002" },
    { "key": "old-key",   "value": "sdk-********-****-****-****-*******00003", "expiry": 1735689600000 }
  ],
  "mobileKey": "mob-********-****-****-****-*******00009",
  "mobileKeys": [
    { "key": "default-mobile", "value": "mob-********-****-****-****-*******00009" }
  ],
  "expiringSdkKey": "sdk-********-****-****-****-*******00003",
  "status": "connected",
  "connectionStatus": { "state": "VALID", "stateSince": 10000000 },
  "dataStoreStatus":  { "state": "VALID", "stateSince": 10000000 }
}
```

## Changes

The rotator exposes `AcceptedKeys()` — the full accepted set as `[]AcceptedKey`, each entry tagged server or mobile via a `KeyType` and carrying the credential value, the optional wire `key` identifier (`*string`, nil when absent), and the optional expiry. The status handler partitions that slice by type into the `sdkKeys[]`/`mobileKeys[]` arrays (obscuring each value) and selects the soonest-expiring non-anchor SDK key for `expiringSdkKey` via `slices.MinFunc`.

On the build side, `AcceptedSetBuilder`'s `WithAnchor` / `WithSDKKey` / `WithMobileKey` / `WithPrimaryMobileKey` take `SDKKeyParams` / `MobileKeyParams` structs carrying the value plus the optional identifier and expiry; `AcceptedSet` stores each as an `acceptedKeyInfo`, and `BuildAcceptedSet` threads the wire `key` through (converting empty → nil via the shared `util.PtrOrNil` helper).

---

> [!NOTE]
> **Medium Risk**
> Changes credential reconcile/build validation and a public status JSON schema; mis-validation could block RAC updates or confuse operators, but behavior is heavily tested and scalars stay backward-compatible.
> 
> **Overview**
> Extends **`/status`** so each environment reports the **full accepted SDK and mobile key sets** via new **`sdkKeys[]`** and **`mobileKeys[]`** arrays (`KeyStatus`: optional wire identifier, obscured value, optional expiry). Legacy scalar **`sdkKey`** / **`mobileKey`** remain and still mark the anchor and primary entries.
> 
> Credential plumbing now carries per-key **wire identifiers** end-to-end: **`AcceptedSetBuilder`** takes **`SDKKeyParams` / `MobileKeyParams`**, **`Rotator.AcceptedKeys()`** exposes metadata for the handler, and reconcile **refreshes or clears** identifiers so `/status` does not show stale names after payload changes. **`BuildAcceptedSet`** rejects a defined **`mobKey`** missing from **`mobileKeys[]`** (parallel to the anchor invariant).
> 
> **`expiringSdkKey`** is now the **soonest-expiring non-anchor** SDK key, with a deterministic tie-break on credential value instead of loop-order luck.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6761994bf1d4acdd253c0f81980e0cb7e85d08b7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/api/status_reps.go
+++ b/internal/api/status_reps.go
@@ -15,17 +15,40 @@ type StatusRep struct {
 	ClientVersion string                          `json:"clientVersion"`
 }
 
+// KeyStatus is the JSON representation of one accepted SDK or mobile key in the status endpoint's
+// sdkKeys[] / mobileKeys[] arrays.
+//
+// Key is the non-secret human-readable identifier from the wire format (the "key" field of a
+// sdkKeys/mobileKeys entry). Value is the obscured credential secret (via sdks.ObscureKey). Expiry
+// carries the Unix-millisecond expiry timestamp when the key is being phased out; it is omitted for
+// permanent keys.
+type KeyStatus struct {
+	Key    string `json:"key"`
+	Value  string `json:"value"`
+	Expiry *int64 `json:"expiry,omitempty"`
+}
+
 // EnvironmentStatusRep is the per-environment JSON representation returned by the status endpoint.
 //
 // This is exported for use in integration test code.
 type EnvironmentStatusRep struct {
-	SDKKey           string               `json:"sdkKey"`
+	// SDKKey is the obscured anchor SDK key — the key relay uses for its upstream connection.
+	// Kept for back-compat; for the full non-anchor accepted set see SDKKeys.
+	SDKKey string `json:"sdkKey"`
+	// SDKKeys carries all non-anchor accepted SDK keys with their identifiers, obscured values, and
+	// optional expiry. Present-but-empty for single-key environments.
+	SDKKeys          []KeyStatus          `json:"sdkKeys"`
 	EnvID            string               `json:"envId,omitempty"`
 	EnvKey           string               `json:"envKey,omitempty"`
 	EnvName          string               `json:"envName,omitempty"`
 	ProjKey          string               `json:"projKey,omitempty"`
 	ProjName         string               `json:"projName,omitempty"`
-	MobileKey        string               `json:"mobileKey,omitempty"`
+	// MobileKey is the obscured primary mobile key. Kept for back-compat; for the full non-primary
+	// accepted set see MobileKeys.
+	MobileKey string `json:"mobileKey,omitempty"`
+	// MobileKeys carries all non-primary accepted mobile keys. Present-but-empty for environments
+	// without additional mobile keys.
+	MobileKeys       []KeyStatus          `json:"mobileKeys"`
 	ExpiringSDKKey   string               `json:"expiringSdkKey,omitempty"`
 	Status           string               `json:"status"`
 	ConnectionStatus ConnectionStatusRep  `json:"connectionStatus"`
--- a/internal/api/status_reps.go
+++ b/internal/api/status_reps.go
@@ -33,8 +33,8 @@ type KeyStatus struct {
 //
 // This is exported for use in integration test code.
 type EnvironmentStatusRep struct {
-	// SDKKey is the obscured anchor SDK key — the key relay uses for its upstream connection. Kept for
-	// back-compat; it designates which SDKKeys entry is the anchor.
+	// SDKKey is the obscured anchor SDK key — the key relay uses for its upstream connection. It
+	// designates which SDKKeys entry is the anchor.
 	SDKKey string `json:"sdkKey"`
 	// SDKKeys carries the full accepted set of server-side SDK keys — including the anchor — with their
 	// identifiers, obscured values, and optional expiry. Always present; always contains at least the
@@ -45,8 +45,7 @@ type EnvironmentStatusRep struct {
 	EnvName  string      `json:"envName,omitempty"`
 	ProjKey  string      `json:"projKey,omitempty"`
 	ProjName string      `json:"projName,omitempty"`
-	// MobileKey is the obscured primary mobile key. Kept for back-compat; it designates which
-	// MobileKeys entry is the primary.
+	// MobileKey is the obscured primary mobile key. It designates which MobileKeys entry is the primary.
 	MobileKey string `json:"mobileKey,omitempty"`
 	// MobileKeys carries the full accepted set of mobile keys — including the primary. Always present;
 	// empty for an environment with no mobile keys (e.g. server-side only).
--- a/internal/api/status_reps.go
+++ b/internal/api/status_reps.go
@@ -19,11 +19,12 @@ type StatusRep struct {
 // sdkKeys[] / mobileKeys[] arrays.
 //
 // Key is the non-secret human-readable identifier from the wire format (the "key" field of a
-// sdkKeys/mobileKeys entry). Value is the obscured credential secret (via sdks.ObscureKey). Expiry
-// carries the Unix-millisecond expiry timestamp when the key is being phased out; it is omitted for
-// permanent keys.
+// sdkKeys/mobileKeys entry); it is omitted when the source carried no identifier (manual config, or
+// an old-format payload predating concurrent keys). Value is the obscured credential secret (via
+// sdks.ObscureKey). Expiry carries the Unix-millisecond expiry timestamp when the key is being phased
+// out; it is omitted for permanent keys.
 type KeyStatus struct {
-	Key    string `json:"key"`
+	Key    string `json:"key,omitempty"`
 	Value  string `json:"value"`
 	Expiry *int64 `json:"expiry,omitempty"`
 }
@@ -32,22 +33,23 @@ type KeyStatus struct {
 //
 // This is exported for use in integration test code.
 type EnvironmentStatusRep struct {
-	// SDKKey is the obscured anchor SDK key — the key relay uses for its upstream connection.
-	// Kept for back-compat; for the full non-anchor accepted set see SDKKeys.
+	// SDKKey is the obscured anchor SDK key — the key relay uses for its upstream connection. Kept for
+	// back-compat; it designates which SDKKeys entry is the anchor.
 	SDKKey string `json:"sdkKey"`
-	// SDKKeys carries all non-anchor accepted SDK keys with their identifiers, obscured values, and
-	// optional expiry. Present-but-empty for single-key environments.
+	// SDKKeys carries the full accepted set of server-side SDK keys — including the anchor — with their
+	// identifiers, obscured values, and optional expiry. Always present; always contains at least the
+	// anchor.
 	SDKKeys  []KeyStatus `json:"sdkKeys"`
 	EnvID    string      `json:"envId,omitempty"`
 	EnvKey   string      `json:"envKey,omitempty"`
 	EnvName  string      `json:"envName,omitempty"`
 	ProjKey  string      `json:"projKey,omitempty"`
 	ProjName string      `json:"projName,omitempty"`
-	// MobileKey is the obscured primary mobile key. Kept for back-compat; for the full non-primary
-	// accepted set see MobileKeys.
+	// MobileKey is the obscured primary mobile key. Kept for back-compat; it designates which
+	// MobileKeys entry is the primary.
 	MobileKey string `json:"mobileKey,omitempty"`
-	// MobileKeys carries all non-primary accepted mobile keys. Present-but-empty for environments
-	// without additional mobile keys.
+	// MobileKeys carries the full accepted set of mobile keys — including the primary. Always present;
+	// empty for an environment with no mobile keys (e.g. server-side only).
 	MobileKeys       []KeyStatus          `json:"mobileKeys"`
 	ExpiringSDKKey   string               `json:"expiringSdkKey,omitempty"`
 	Status           string               `json:"status"`
--- a/internal/api/status_reps.go
+++ b/internal/api/status_reps.go
@@ -37,12 +37,12 @@ type EnvironmentStatusRep struct {
 	SDKKey string `json:"sdkKey"`
 	// SDKKeys carries all non-anchor accepted SDK keys with their identifiers, obscured values, and
 	// optional expiry. Present-but-empty for single-key environments.
-	SDKKeys          []KeyStatus          `json:"sdkKeys"`
-	EnvID            string               `json:"envId,omitempty"`
-	EnvKey           string               `json:"envKey,omitempty"`
-	EnvName          string               `json:"envName,omitempty"`
-	ProjKey          string               `json:"projKey,omitempty"`
-	ProjName         string               `json:"projName,omitempty"`
+	SDKKeys  []KeyStatus `json:"sdkKeys"`
+	EnvID    string      `json:"envId,omitempty"`
+	EnvKey   string      `json:"envKey,omitempty"`
+	EnvName  string      `json:"envName,omitempty"`
+	ProjKey  string      `json:"projKey,omitempty"`
+	ProjName string      `json:"projName,omitempty"`
 	// MobileKey is the obscured primary mobile key. Kept for back-compat; for the full non-primary
 	// accepted set see MobileKeys.
 	MobileKey string `json:"mobileKey,omitempty"`
--- a/internal/credential/accepted_set.go
+++ b/internal/credential/accepted_set.go
@@ -35,6 +35,12 @@ type AcceptedSet struct {
 	mobileKeys       map[config.MobileKey]*time.Time
 	primaryMobileKey config.MobileKey
 	envID            config.EnvironmentID
+
+	// sdkKeyIdentifiers and mobileKeyIdentifiers carry the wire "key" field — the non-secret
+	// human-readable identifier for each credential. May be absent for entries synthesized from
+	// old-format payloads that only have the singular sdkKey/mobKey fields.
+	sdkKeyIdentifiers    map[config.SDKKey]string
+	mobileKeyIdentifiers map[config.MobileKey]string
 }
 
 // hasSDKKey reports whether key is one of the set's accepted SDK keys.
--- a/internal/credential/accepted_set.go
+++ b/internal/credential/accepted_set.go
@@ -18,8 +18,8 @@ import (
 //
 // WithAnchor / WithPrimaryMobileKey both add the key to the set and designate it, so adding a
 // single key takes one call. Build requires that an anchor was designated. (Structural validation of
-// the wire payload — undefined credentials, an anchor absent from the array — happens upstream when
-// the payload is parsed into the set; see SDK-2547.)
+// the wire payload — undefined credentials, a designated key absent from its array — happens upstream
+// when the payload is parsed into the set.)
 //
 // A key's expiry is taken from its entry in this set; the legacy sdkKey.expiring{} wire slot is not
 // consulted when building it.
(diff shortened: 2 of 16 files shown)


internal/credential/accepted_set.go
```diff
@@ -35,6 +35,12 @@ type AcceptedSet struct {
 	mobileKeys       map[config.MobileKey]*time.Time
 	primaryMobileKey config.MobileKey
 	envID            config.EnvironmentID
+
+	// sdkKeyIdentifiers and mobileKeyIdentifiers carry the wire "key" field — the non-secret
+	// human-readable identifier for each credential. May be absent for entries synthesized from
+	// old-format payloads that only have the singular sdkKey/mobKey fields.
+	sdkKeyIdentifiers    map[config.SDKKey]string
+	mobileKeyIdentifiers map[config.MobileKey]string
```
aaron-zeisler: I think the additions to the AcceptedSet's properties and the AcceptedSetBuilder's API can be improved.

Instead of adding this new set of maps, I think we should define a struct that represents the key, similar to what's been defined in the Rotator. If the key (ie identifier) and expiration are really optional, those properties of the struct can be pointers. Please dig deep to understand in the key (ie identifier) is really optional.

Then we don't need the new WithSDKKeyIdentifier and WithMobileKeyIdentifier functions on the builder. The existing "With..." functions can take the struct as a parameter instead.

aaron-zeisler:
Throughout this pull request, "Identifier" should be called "Key". Per Matthew's comment here: [[link](https://github.com/launchdarkly/ld-relay/pull/702#discussion_r3430116856)]. I understand the design doc may call it "Identifier" - let's use "Key" everywhere, and update the design/plan document to reflect that.

internal/credential/rotator.go
```diff
@@ -10,7 +11,27 @@ import (
...
+	Expiry *time.Time
+}
+
+// MobileKeyEntry is metadata for one accepted mobile key, returned by NonPrimaryMobileKeyEntries.
+// Used by the status endpoint to populate the mobileKeys[] response array.
+type MobileKeyEntry struct {
+	Value      config.MobileKey
+	Identifier string
+	Expiry     *time.Time
 }
```
aaron-zeisler: Both structs should have struct-level and property-level comments

internal/credential/rotator.go
```diff
@@ -533,3 +565,115 @@ func (r *Rotator) reconcileEnvironmentID(set AcceptedSet) {
...
+			if ai == "" {
+				return false
+			}
+			if bi == "" {
+				return true
+			}
+			return ai < bi
+		}
+		return value(entries[i]) < value(entries[j])
+	})
```
aaron-zeisler: Does the result really need to be sorted?

internal/credential/rotator.go
```diff
@@ -533,3 +565,115 @@ func (r *Rotator) reconcileEnvironmentID(set AcceptedSet) {
...
+// identifier support) sort after identified keys; ties within no-identifier keys are broken by
+// credential value for determinism.
+func nonPrimaryKeyEntries[K ~string, E any](
+	accepted map[K]*acceptedKeyInfo,
+	deprecated map[K]time.Time,
+	primary K,
+	identifier func(E) string,
+	value func(E) string,
+	construct func(key K, identifier string, expiry *time.Time) E,
+) []E {
```
aaron-zeisler: This helper function's interface is very heavy and seems like overkill. What are the other options to achieve the same thing? Is this entire function even necessary?

At this level of the service, is it really necessary for SdkKeyEntry and MobileKeyEntry are different structs? Can they be the same struct? What are the trade-offs for keeping them separate?

internal/credential/rotator.go
```diff
@@ -533,3 +565,115 @@ func (r *Rotator) reconcileEnvironmentID(set AcceptedSet) {
...
+			updateEarliest(key, *info.expiry)
+		}
+	}
+	// Legacy RotateWithGrace path: expiry stored in deprecatedSdkKeys.
+	for key, t := range r.deprecatedSdkKeys {
+		updateEarliest(key, t)
+	}
+
+	return earliest, earliestTime != nil
+}
```
aaron-zeisler, line 362: Is there any performance impact if we just sort the list using `slices.SortFunc()` and then take the 0th index?

aaron-zeisler: Looking at the rest of the code, I'm wondering if this function is necessary at all. Since it's only used in the status endpoint handler, Could the handler code simply call NonAnchorSDKKeyEntries or GetAcceptedSdkKeys, and then find the earliest based on the simple algorithm in my "performance impact" comment above?

internal/api/status_reps.go
```diff
@@ -15,17 +15,42 @@ type StatusRep struct {
...
 type EnvironmentStatusRep struct {
-	SDKKey           string               `json:"sdkKey"`
-	EnvID            string               `json:"envId,omitempty"`
-	EnvKey           string               `json:"envKey,omitempty"`
-	EnvName          string               `json:"envName,omitempty"`
-	ProjKey          string               `json:"projKey,omitempty"`
-	ProjName         string               `json:"projName,omitempty"`
-	MobileKey        string               `json:"mobileKey,omitempty"`
+	// SDKKey is the obscured anchor SDK key — the key relay uses for its upstream connection. Kept for
+	// back-compat; it designates which SDKKeys entry is the anchor.
```
aaron-zeisler: Here and in the comment below, remove the detail about backwards-compatibility: the comments should reflect the current state of the code, and should not describe the history of the code.

internal/credential/accepted_set.go
```diff
@@ -27,16 +27,24 @@ import (
...
-	mobileKeys       map[config.MobileKey]*time.Time
+	mobileKeys       map[config.MobileKey]acceptedKeyMeta
 	primaryMobileKey config.MobileKey
 	envID            config.EnvironmentID
 }
 
+// acceptedKeyMeta is the per-key metadata stored in an AcceptedSet alongside the credential value:
+// the optional wire "key" identifier (nil when the source carried none — manual config or an
+// old-format payload) and the optional expiry (nil = permanent).
+type acceptedKeyMeta struct {
```
aaron-zeisler: Let's call this acceptedKeyMetadata

internal/credential/rotator.go
```diff
@@ -11,6 +11,46 @@ import (
...
+	Type KeyType
+	// Value is the credential secret (what the wire calls "value"). Plain, not obscured; the status
+	// handler obscures it before serialising.
+	Value string
+	// Key is the non-secret human-readable identifier (the wire "key" field). Nil when the source
+	// carried no identifier — manual configuration, or an old-format payload predating concurrent keys.
+	Key *string
+	// Expiry is the key's expiry time. Nil means the key is permanent.
+	Expiry *time.Time
 }
```
aaron-zeisler, line 41: In this package, we now have AcceptedKey, acceptedKeyInfo, SDKKeyParams, and acceptedKeyMeta. Let's reduce this to hopefully just 1 or possibly 2 structs.

internal/credential/accepted_set_builder_test.go
```diff
@@ -52,3 +53,24 @@ func mustBuild(t *testing.T, b *AcceptedSetBuilder) AcceptedSet {
...
+}
+func sdkPExp(value config.SDKKey, key string, expiry time.Time) SDKKeyParams {
+	return SDKKeyParams{Value: value, Key: strPtrOrNil(key), Expiry: &expiry}
+}
+func mobP(value config.MobileKey, key string) MobileKeyParams {
+	return MobileKeyParams{Value: value, Key: strPtrOrNil(key)}
+}
+func mobPExp(value config.MobileKey, key string, expiry time.Time) MobileKeyParams {
+	return MobileKeyParams{Value: value, Key: strPtrOrNil(key), Expiry: &expiry}
+}
```
aaron-zeisler: Let's not have these 4 helper functions. I find the code easier to read without these functions. strPtrOrNil is fine.

internal/envfactory/reconcile_helper.go
```diff
@@ -1,10 +1,29 @@
...
+}
+
+// expiryPtr returns a pointer to the expiry time, or nil when it is the zero time (a permanent key).
+func expiryPtr(expiry time.Time) *time.Time {
+	if expiry.IsZero() {
+		return nil
+	}
+	return &expiry
+}
+
```
aaron-zeisler: Does this repository, one of LaunchDarkly's other repositories (go-sdk-common, etc...), or one of the libraries that this service imports have access to a set of generic helper functions that perform these actions? It seems like these could be generalized to StringPtrOrNil and TimePtrOrNil if they were in a library of utility functions.

internal/envfactory/reconcile_helper_test.go
```diff
@@ -24,6 +24,21 @@ func mustBuild(t *testing.T, b *credential.AcceptedSetBuilder) credential.Accept
...
+func sdkPExp(value config.SDKKey, key string, expiry time.Time) credential.SDKKeyParams {
+	return credential.SDKKeyParams{Value: value, Key: keyPtr(key), Expiry: expiryPtr(expiry)}
+}
+func mobP(value config.MobileKey, key string) credential.MobileKeyParams {
+	return credential.MobileKeyParams{Value: value, Key: keyPtr(key)}
+}
+func mobPExp(value config.MobileKey, key string, expiry time.Time) credential.MobileKeyParams {
+	return credential.MobileKeyParams{Value: value, Key: keyPtr(key), Expiry: expiryPtr(expiry)}
+}
+
```
aaron-zeisler: Again, let's not include these 4 functions. The code is easier to read without them.

relay/endpoints_status.go
```diff
@@ -155,3 +187,17 @@ func statusHandler(relay *Relay) http.Handler {
...
+// secret value and surfacing the optional identifier and expiry.
+func keyStatus(k credential.AcceptedKey) api.KeyStatus {
+	ks := api.KeyStatus{Value: sdks.ObscureKey(k.Value)}
+	if k.Key != nil {
+		ks.Key = *k.Key
+	}
+	if k.Expiry != nil {
+		ms := k.Expiry.UnixMilli()
+		ks.Expiry = &ms
+	}
```
aaron-zeisler, line 204: If expiration is expressed in milliseconds when it's ingested by Relay Proxy, and it's also expressed in milliseconds in the status handler's response, why convert it to time.Time within the service's code? What are the benefits to the conversion?

internal/relayenv/env_context_impl_test.go
```diff
@@ -187,6 +187,9 @@ func mustBuildAcceptedSet(t *testing.T, b *credential.AcceptedSetBuilder) creden
 	return set
 }
 
+// expiryPtr returns a pointer to t, for building credential param structs in these tests.
+func expiryPtr(t time.Time) *time.Time { return &t }
+
```
aaron-zeisler: Can this be replaced by PtrOrNil?

internal/envfactory/reconcile_helper.go
```diff
@@ -50,18 +44,15 @@ func BuildAcceptedSet(params EnvironmentParams) (credential.AcceptedSet, config.
...
-			b.WithExpiringSDKKey(k.Value, k.Expiry)
+			b.WithSDKKey(credential.SDKKeyParams{Value: k.Value, Key: util.PtrOrNil(k.Key), Expiry: util.PtrOrNil(k.Expiry)})
 		}
 	}
 
 	// The anchor must be one of the accepted SDK keys: the backend lists it in sdkKeys[] (and ToParams
 	// synthesizes it into the array for old-format payloads). A defined anchor absent from the array is
-	// a structurally malformed payload — reject it per §9 rather than letting WithAnchor above
-	// silently synthesize it into the set.
+	// a structurally malformed payload — reject it per §9.
```
aaron-zeisler: Here's a reference to the project-specific design doc. Please remove project-specific references from comments throughout this PR's code.

relay/endpoints_status.go
```diff
@@ -60,11 +60,41 @@ func statusHandler(relay *Relay) http.Handler {
...
 			}
+			// Arrays are always present (never null): a server-only env has an empty mobileKeys.
+			if status.SDKKeys == nil {
+				status.SDKKeys = []api.KeyStatus{}
+			}
+			if status.MobileKeys == nil {
+				status.MobileKeys = []api.KeyStatus{}
+			}
+
+			// expiringSdkKey (back-compat): the soonest-expiring non-anchor SDK key. Picking the soonest
```
aaron-zeisler: Remove notes about backwards-compatibility: the comments should refer to the code's current state, and not the code's history.

internal/envfactory/reconcile_helper.go
```diff
@@ -70,10 +61,10 @@ func BuildAcceptedSet(params EnvironmentParams) (credential.AcceptedSet, config.
 		if !k.Value.Defined() {
 			return credential.AcceptedSet{}, anchor, credential.NewEmptyCredentialError("mobileKeys", k.Key)
 		}
-		if k.Expiry.IsZero() {
-			b.WithMobileKey(k.Value)
+		if k.Value == params.MobileKey {
+			b.WithPrimaryMobileKey(credential.MobileKeyParams{Value: k.Value, Key: util.PtrOrNil(k.Key)})
```
aaron-zeisler, line 70: Question for here where the primary mobile key is added, and also above on line 47 where the anchor is added: is it possible for the primary mobile key and/or the anchor to have an expiration time? I don't see it in the example payloads, but I assume it's possible.

internal/envfactory/reconcile_helper.go
```diff
@@ -70,10 +61,10 @@ func BuildAcceptedSet(params EnvironmentParams) (credential.AcceptedSet, config.
...
 		} else {
-			b.WithExpiringMobileKey(k.Value, k.Expiry)
+			b.WithMobileKey(credential.MobileKeyParams{Value: k.Value, Key: util.PtrOrNil(k.Key), Expiry: util.PtrOrNil(k.Expiry)})
```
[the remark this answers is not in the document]
aaron-zeisler: I'd like verification on this, but I think it's a rule that the primary/anchor keys must appear in the sdkKeys[] and mobileKeys[] arrays, and I also believe there's validation for this in the Builder. Ah, I see the validation for the anchor, but I don't see validation for the primary mobile key.

internal/credential/rotator.go
```diff
@@ -258,14 +298,20 @@ func reconcileAcceptedKeys[K reconcilableKey](
 // payload may carry for it. The caller must hold the write lock.
 func (r *Rotator) reconcileSDKKeys(set AcceptedSet, anchor config.SDKKey, now time.Time) {
 	desired := make(map[config.SDKKey]*time.Time, len(set.sdkKeys))
-	for key, expiry := range set.sdkKeys {
-		if expiry != nil && !now.Before(*expiry) {
+	for key, meta := range set.sdkKeys {
+		if meta.expiry != nil && !now.Before(*meta.expiry) {
```
aaron-zeisler: Nit: please replace "meta" with "info" throughout this file when looping through set.sdkKeys or set.mobileKeys. I believe "meta" refers to the old name for the struct that was being processed - the struct that's now being processed is called "acceptedKeyInfo".

PR #725, opened 2026-06-27
test(concurrent-keys): integration tests for multi-key downstream auth

aaron-zeisler:
## Summary

Adds integration test coverage for the SDK-connection / authentication scenarios of **concurrent SDK keys** — multi-key environments delivered via the new `sdkKeys[]` / `mobileKeys[]` array wire format — on both the RAC and offline-mode paths. This is the connection/auth slice of the M2 catch-up round.

Jira: [SDK-2609](https://launchdarkly.atlassian.net/browse/SDK-2609)

## Background

The multi-key behavior shipped with per-task **unit** tests rather than the harness-based integration scenarios the plan intended, leaving the new array-format auth path with essentially no end-to-end coverage (two recent regressions were caught by Bugbot, not tests). These tests close that gap for the connection/auth scenarios specifically.

They deliberately do **not** duplicate existing coverage:

- Single-key / legacy `expiring{}`-slot rotation is already covered (`TestAutoConfigInitWithExpiringSDKKey`, `TestOfflineModeSDKKeyCanExpire`, etc.) via the credential **set**; these assert downstream **authentication** and live stream connect/disconnect via the **array** path.
- Generic unknown-credential → 401 is already covered at the middleware / end-to-end layer; here only the multi-key-specific rejection cases are added.

## Changes

New file `relay/concurrent_keys_auth_test.go`:

- **Non-anchor keys authenticate** (RAC + offline) — every accepted SDK and mobile key authenticates downstream, with a single upstream connection (the anchor) serving all.
- **Disconnect on expiry** (offline live + RAC auth) — a connected downstream stream is torn down when its non-anchor key expires; sibling keys stay connected; the expired key then returns 401. Covers an SDK key and a mobile key.
- **Survives future expiry** (offline live + RAC auth) — a connected non-anchor key that gains a not-yet-reached expiry keeps its open stream and keeps authenticating.
- **Per-credential rejection** (RAC + offline) — a credential outside a populated accepted set, and a key removed on reconcile, are rejected (401) while siblings continue to work.

Notes for review:

- Built on the existing `autoConfTest` / `offlineModeTest` harnesses, fed array-format `EnvironmentRep`s.
- Live open-connection teardown is asserted on the **offline** path (real SDK client serves stream data). The RAC equivalents use `FakeLDClient`, which doesn't serve stream bodies, so they assert the same expiry → reject / future-expiry → still-accepted outcome at the auth layer. Documented inline.
- Expiry tests connect on a *permanent* key first, then introduce the expiry, so the connection establishes independent of expiry timing.
- Authored ahead of SDK-2604 (anchor rename) and SDK-2603 (legacy-path removal). The tests reference no symbols those PRs rename, so a near-zero rebase is expected once they land.

[SDK-2609]: https://launchdarkly.atlassian.net/browse/SDK-2609?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Low Risk**
> Test-only change with no production code paths modified; risk is limited to CI runtime and test maintenance.
> 
> **Overview**
> Adds **`relay/concurrent_keys_auth_test.go`**, a dedicated integration suite for **multi-key** environments (`sdkKeys[]` / `mobileKeys[]`) where one **anchor** owns the single upstream connection.
> 
> Coverage spans **RAC** (`autoConfTest`) and **offline** (`offlineModeTest`): non-anchor SDK/mobile keys authenticate while only the anchor creates an upstream client; credentials outside the accepted set or removed on reconcile are rejected; non-anchor key expiry triggers cleanup, 401, and (offline) live SSE disconnect on `/all` and mobile streams; future expiry keeps streams open and auth working; anchor rotation swaps the upstream client and drops the old anchor while non-anchor keys stay accepted (plus a dummy-client RAC test that a non-anchor stream stays open during rotation).
> 
> Shared helpers build array-format env reps, assert stream open/close, and wait for credential removal from routing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c9e85b8a0816d8d9f45726cbfd1f109a11a255a6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/relay/concurrent_keys_auth_test.go
+++ b/relay/concurrent_keys_auth_test.go
@@ -1,19 +1,18 @@
 package relay
 
-// Integration scenarios for multi-key (concurrent SDK keys) downstream authentication.
+// Downstream authentication for environments that accept multiple SDK keys and multiple mobile keys
+// (the sdkKeys[]/mobileKeys[] array wire format), where a single anchor key owns the one upstream
+// connection. These are permanent regression tests for that behavior.
 //
-// These tests exercise the NEW behavior: an environment whose accepted set carries multiple SDK
-// keys and multiple mobile keys via the sdkKeys[]/mobileKeys[] array wire format, with a single
-// anchor key owning the one upstream connection. They complement — and deliberately do not
-// duplicate — existing coverage:
+// They complement — and deliberately do not duplicate — existing coverage:
 //
 //   - Single-key / legacy expiring{}-slot rotation is covered by TestAutoConfigInitWithExpiringSDKKey,
 //     TestOfflineModeDeprecatedSDKKeyIsRespectedIfExpiryInFuture, TestOfflineModeSDKKeyCanExpire, etc.
-//     Those assert on the credential SET via the legacy rotation path; these assert downstream
-//     AUTHENTICATION (and live stream connect/disconnect) via the array path.
+//     Those assert on the credential set via the legacy rotation path; these assert downstream
+//     authentication (and live stream connect/disconnect) via the array path.
 //   - Generic unknown-credential -> 401 is covered at the middleware/end-to-end layer; here we only
-//     test the multi-key-specific rejection (a credential outside a populated accepted set, and a
-//     key removed from the accepted set on reconcile).
+//     test rejection that is specific to a multi-key environment (a credential outside a populated
+//     accepted set, and a key removed from the accepted set on reconcile).
 //
 // The tests reuse the in-process harnesses (autoConfTest for RAC, offlineModeTest for offline) so
 // they get assertSDKEndpointsAvailability (auth accept/reject) and awaitClient/shouldNotCreateClient
@@ -23,7 +22,7 @@ import (
 	"testing"
 	"time"
 
-	c "github.com/launchdarkly/ld-relay/v8/config"
+	"github.com/launchdarkly/ld-relay/v8/config"
 	"github.com/launchdarkly/ld-relay/v8/internal/credential"
 	"github.com/launchdarkly/ld-relay/v8/internal/envfactory"
 	"github.com/launchdarkly/ld-relay/v8/internal/filedata"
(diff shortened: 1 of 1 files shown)


relay/concurrent_keys_auth_test.go
```diff
@@ -0,0 +1,433 @@
...
+//
+// The tests reuse the in-process harnesses (autoConfTest for RAC, offlineModeTest for offline) so
+// they get assertSDKEndpointsAvailability (auth accept/reject) and awaitClient/shouldNotCreateClient
+// (proving the anchor owns the only upstream connection).
+
+import (
+	"testing"
+	"time"
+
+	c "github.com/launchdarkly/ld-relay/v8/config"
```
aaron-zeisler: I don't think there's a good reason to alias this import.

relay/concurrent_keys_auth_test.go
```diff
@@ -0,0 +1,433 @@
...
+	ckExtraMob  = c.MobileKey("mob-ck-extra")
+	ckFlagKey   = "ck-flag"
+)
+
+var ckIdentifiers = relayenv.EnvIdentifiers{
+	ProjName: "CK Project",
+	ProjKey:  "ck-proj",
+	EnvName:  "CK Env",
+	EnvKey:   "ck-env",
+}
```
aaron-zeisler, line 67: I think these tests should be useful after the concurrent-keys project is complete. Is the idea to remove these tests? If not, let's use names that are less project-specific.

relay/concurrent_keys_auth_test.go
```diff
@@ -0,0 +1,433 @@
...
+		{
+			Kind: ldstoreimpl.Features(),
+			Items: []ldstoretypes.KeyedItemDescriptor{
+				{Key: ckFlagKey, Item: sharedtest.FlagDesc(flag)},
+			},
+		},
+	}
+}
+
+// --- S5: a valid non-anchor key authenticates downstream; one upstream connection serves all ---
```
aaron-zeisler: I believe "S5" is a project-specific reference. Let's not include project-specific references within these tests.

relay/concurrent_keys_auth_test.go
```diff
@@ -0,0 +1,433 @@
...
+// key the singular sdkKey.value/mobKey points to; the "extra" keys are additional accepted entries
+// that are NOT the anchor.
+const (
+	ckEnvID     = c.EnvironmentID("ck-multikey-env")
+	ckAnchorSDK = c.SDKKey("sdk-ck-anchor")
+	ckExtraSDK  = c.SDKKey("sdk-ck-extra")
+	ckAnchorMob = c.MobileKey("mob-ck-anchor")
+	ckExtraMob  = c.MobileKey("mob-ck-extra")
+	ckFlagKey   = "ck-flag"
+)
```
aaron-zeisler, line 60: In general, I'm not a fan of the "ck" prefix that's added to variable and function names throughout this file. Can we expand that abbreviation? Is it more suitable to put these tests in their own package (to limit the exposure of these function from the rest of the relay package, and so the naming can be less verbose)?

relay/concurrent_keys_auth_test.go
```diff
@@ -0,0 +1,432 @@
...
+			} else {
+				p.assertSDKEndpointsAvailability(false, "", extraMobileKey, "")
+			}
+			p.assertSDKEndpointsAvailability(true, anchorSDKKey, anchorMobileKey, multiKeyEnvID)
+		})
+	}
+
+	t.Run("sdk key", func(t *testing.T) { run(t, "/all", "put", extraSDKKey, true) })
+	t.Run("mobile key", func(t *testing.T) {
+		run(t, "/meval/eyJrZXkiOiJ1c2Vya2V5In0=", "ping", extraMobileKey, false)
```
keelerm84: The example you provided is for the legacy user format. Better to have a valid context format here.

```suggestion
		run(t, "/meval/eyJrZXkiOiJ1c2Vya2V5IiwgImtpbmQiOiJ1c2VyIn0==", "ping", extraMobileKey, false)
```

keelerm84 (approved):
You should definitely expand these to cover rotating out the anchor key, and also ensuring a non-anchor key stays connected during anchor key rotation.

relay/concurrent_keys_auth_test.go
aaron-zeisler: Thanks! I've updated this specific line. I see that this format is also in a utility function called simpleUserBase64 - I've kicked off a separate PR to update the shared function as well (linked below).

aaron-zeisler:
> You should definitely expand these to cover rotating out the anchor key, and also ensuring a non-anchor key stays connected during anchor key rotation.

Thanks, I've added these two tests. I have a list of additional tests and edge cases that I'll continue to add in chunks.

PR #726, opened 2026-06-29
chore(credential): remove DeprecatedCredentials / GetDeprecatedCredentials

aaron-zeisler:
## Summary
Remove `Rotator.DeprecatedCredentials()` and `EnvContext.GetDeprecatedCredentials()` — both methods are now dead code following PR #724.

## Background
PR #724 (T4 — status endpoint arrays) rewired the `/status` handler to compute the `expiringSdkKey` field from `GetAcceptedKeys()` instead of `GetDeprecatedCredentials()`. After that change, `GetDeprecatedCredentials` had no production callers — only tests referenced it.

Tests that asserted rotation/expiry state through the deleted methods are rewritten to filter `AcceptedKeys()` / `GetAcceptedKeys()` for non-anchor SDK keys with a non-nil expiry — the same semantics, expressed through a surviving accessor.

---

> [!NOTE]
> **Low Risk**
> Dead-code removal with no remaining references; credential rotation and auth behavior are unchanged aside from dropping unused accessors.
> 
> **Overview**
> Removes **`Rotator.DeprecatedCredentials()`**, **`EnvContext.GetDeprecatedCredentials()`**, and the **`envContextImpl`** passthrough — APIs that had no production callers after `/status` started deriving expiring SDK keys from **`GetAcceptedKeys()`**.
> 
> Tests that previously depended on those methods now use a local **`expiringSDKKeys`** helper (on the rotator, env context, and relay test packages) that returns non-anchor accepted server SDK keys with a non-nil expiry — same semantics as the deleted methods, without keeping the public surface.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c785fecadbec878c10781598655308f3523432b6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/credential/rotator_test.go
+++ b/internal/credential/rotator_test.go
@@ -61,7 +61,7 @@ func TestReconcileAnchorOnly(t *testing.T) {
 	assert.Empty(t, expirations)
 	assert.Equal(t, anchor, r.AnchorKey())
 	assert.ElementsMatch(t, []SDKCredential{anchor}, r.AllCredentials())
-	assert.Empty(t, r.DeprecatedCredentials())
+	assert.Empty(t, expiringSDKKeys(r))
 }
 
 func TestReconcileMultipleSDKKeys(t *testing.T) {
@@ -79,7 +79,7 @@ func TestReconcileMultipleSDKKeys(t *testing.T) {
 	assert.Empty(t, expirations)
 	assert.Equal(t, anchor, r.AnchorKey())
 	assert.ElementsMatch(t, []SDKCredential{anchor, other}, r.AllCredentials())
-	assert.Empty(t, r.DeprecatedCredentials())
+	assert.Empty(t, expiringSDKKeys(r))
 }
 
 func TestReconcileMultipleMobileKeys(t *testing.T) {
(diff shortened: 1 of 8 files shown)


internal/credential/rotator_test.go
```diff
@@ -259,14 +259,28 @@ func TestReconcileDeExpiryRestoresKey(t *testing.T) {
...
 	additions, expirations := r.StepTime(expiry.Add(1 * time.Millisecond))
 	assert.Empty(t, additions)
 	assert.Empty(t, expirations)
 	assert.Contains(t, r.AllCredentials(), SDKCredential(key))
 }
 
+// expiringSDKKeys returns the non-anchor accepted SDK keys that carry a non-nil expiry. This is the
+// set that DeprecatedCredentials used to return before it was removed; tests use it to assert
+// rotation/expiry state without depending on that deleted method.
+func expiringSDKKeys(r *Rotator) []SDKCredential {
```
keelerm84, line 273: You have the same function defined like 3 different places in this PR. Is there any room for re-use here?

aaron-zeisler:
Looking at the tests, it seems like this GetDeprecatedCredentials() function provides value throughout the layer of the service, even if it isn't called in production code any more. I've closed the PR in favor of keeping it.

PR #729, opened 2026-06-29
refactor(credential): carry wire key identifiers through the accepted set

aaron-zeisler:
> [!NOTE]
> **Stacked PR — splitting #724 (review/merge bottom-up):**
> 1. #728 — reject primary mobile key absent from `mobileKeys[]`
> 2. **➡ #729 (this PR)** — carry wire key identifiers through the accepted set
> 3. #730 — expose the full accepted key set via `AcceptedKeys`
> 4. #731 — surface `sdkKeys[]` / `mobileKeys[]` on `/status`
>
> Base: #728.

## Summary

Carry the wire `key` identifier through the credential accepted-set model so it can later be surfaced on the `/status` endpoint. This is the storage/model layer only — no externally visible behaviour changes yet.

Jira: [SDK-2534](https://launchdarkly.atlassian.net/browse/SDK-2534)

## Background

Second in the stack splitting #724, based on #728. Each accepted key needs to carry its optional non-secret identifier alongside its value and expiry.

Because this PR sits on #728, the per-entry primary-mobile designation introduced here is paired with #728's `mobKey`-not-in-`mobileKeys[]` rejection: a defined primary that is absent from the array is rejected outright rather than silently left undesignated.

## Changes

- Replace the builder's `WithExpiring{SDK,Mobile}Key` with a param-struct API (`SDKKeyParams` / `MobileKeyParams`).
- `AcceptedSet` map values become `acceptedKeyInfo` (value, expiry, optional `key`).
- Rotator reconcile loops thread the identifier through, refreshing it each reconcile and clearing it when a later payload carries none.
- Add `util.PtrOrNil` for optional-field modelling.

---

> [!NOTE]
> **Medium Risk**
> Changes the credential reconcile/build path that gates which keys are accepted, but authentication semantics are unchanged; the main new behavior is metadata refresh on reconcile (identifiers can clear or update).
> 
> **Overview**
> Extends the credential **accepted-set** model so each SDK/mobile key stores optional wire **`key`** identifiers (plus expiry) alongside the secret value, preparing later **`/status`** exposure without changing auth behavior in this PR.
> 
> The **`AcceptedSetBuilder`** API is consolidated: separate expiring-key helpers are replaced by **`SDKKeyParams`** / **`MobileKeyParams`** (value, optional identifier, optional expiry). **`WithAnchor`** / **`WithPrimaryMobileKey`** still force permanent keys and can overwrite earlier entries for the same value.
> 
> **`BuildAcceptedSet`** now designates anchor and primary mobile keys while walking the wire arrays and passes identifiers through **`util.PtrOrNil`**. Rename tests expect **`AcceptedSet`** equality to differ when only the display identifier changes.
> 
> The **rotator** stores **`acceptedKeyInfo`** by value and, on reconcile, **replaces** full per-key metadata so identifiers refresh and absent wire **`key`** fields clear stale names. Reconcile no longer re-injects anchor/primary into the desired map separately; it trusts the built set from the builder.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5eceeb4a8b16c2fa2f682d2c20e9d24f2481d6a3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/credential/accepted_set_builder.go
+++ b/internal/credential/accepted_set_builder.go
@@ -16,77 +16,69 @@ type AcceptedSetBuilder struct {
 func NewAcceptedSetBuilder() *AcceptedSetBuilder {
 	return &AcceptedSetBuilder{
 		set: AcceptedSet{
-			sdkKeys:    make(map[config.SDKKey]*time.Time),
-			mobileKeys: make(map[config.MobileKey]*time.Time),
+			sdkKeys:    make(map[config.SDKKey]acceptedKeyInfo),
+			mobileKeys: make(map[config.MobileKey]acceptedKeyInfo),
 		},
 	}
 }
 
-// WithSDKKey adds a permanent (non-expiring) SDK key. It is a no-op if the key is undefined or
-// already present.
-func (b *AcceptedSetBuilder) WithSDKKey(key config.SDKKey) *AcceptedSetBuilder {
-	b.addSDKKey(key, nil)
-	return b
+// SDKKeyParams describes one accepted server-side SDK key for the builder: the credential value plus
+// the optional wire "key" identifier (nil when absent) and optional expiry (nil = permanent).
+type SDKKeyParams struct {
+	Value  config.SDKKey
+	Key    *string
+	Expiry *time.Time
 }
 
-// WithExpiringSDKKey adds an SDK key that should be accepted until the given expiry. It is a no-op
-// if the key is undefined or already present.
-func (b *AcceptedSetBuilder) WithExpiringSDKKey(key config.SDKKey, expiry time.Time) *AcceptedSetBuilder {
-	b.addSDKKey(key, &expiry)
-	return b
+// MobileKeyParams describes one accepted mobile key for the builder. See SDKKeyParams.
+type MobileKeyParams struct {
+	Value  config.MobileKey
+	Key    *string
+	Expiry *time.Time
 }
 
-// WithAnchor adds key (if not already present) and designates it as the anchor — the SDK key
-// that owns the environment's upstream connection. It is a no-op if the key is undefined.
-func (b *AcceptedSetBuilder) WithAnchor(key config.SDKKey) *AcceptedSetBuilder {
-	if key.Defined() {
-		b.addSDKKey(key, nil)
-		b.set.anchor = key
+// WithSDKKey adds a server-side SDK key. It is a no-op if the value is undefined or already present
+// (the first metadata recorded for a value wins).
+func (b *AcceptedSetBuilder) WithSDKKey(p SDKKeyParams) *AcceptedSetBuilder {
+	if !p.Value.Defined() || b.set.hasSDKKey(p.Value) {
+		return b
 	}
+	b.set.sdkKeys[p.Value] = acceptedKeyInfo{key: p.Key, expiry: p.Expiry}
 	return b
 }
 
-// addSDKKey records the key with the given expiry (nil = permanent), skipping undefined keys and
-// keys already in the set (the first expiry recorded for a key wins).
-func (b *AcceptedSetBuilder) addSDKKey(key config.SDKKey, expiry *time.Time) {
-	if !key.Defined() || b.set.hasSDKKey(key) {
-		return
+// WithAnchor adds p.Value and designates it as the anchor — the SDK key that owns the environment's
+// upstream connection. The anchor is always permanent, so p.Expiry is ignored. It is a no-op if the
+// value is undefined. Unlike WithSDKKey it overwrites any existing entry for the value, since
+// designating the anchor takes precedence over an earlier non-anchor add.
+func (b *AcceptedSetBuilder) WithAnchor(p SDKKeyParams) *AcceptedSetBuilder {
+	if !p.Value.Defined() {
+		return b
 	}
-	b.set.sdkKeys[key] = expiry
-}
-
-// WithMobileKey adds a permanent (non-expiring) mobile key. It is a no-op if the key is undefined or
-// already present.
-func (b *AcceptedSetBuilder) WithMobileKey(key config.MobileKey) *AcceptedSetBuilder {
-	b.addMobileKey(key, nil)
+	b.set.sdkKeys[p.Value] = acceptedKeyInfo{key: p.Key, expiry: nil}
+	b.set.anchor = p.Value
 	return b
 }
 
-// WithExpiringMobileKey adds a mobile key that should be accepted until the given expiry. It is a
-// no-op if the key is undefined or already present.
-func (b *AcceptedSetBuilder) WithExpiringMobileKey(key config.MobileKey, expiry time.Time) *AcceptedSetBuilder {
-	b.addMobileKey(key, &expiry)
-	return b
-}
-
-// WithPrimaryMobileKey adds key (if not already present) and designates it as the primary mobile
-// key — the singular default (the wire's mobKey) used where one mobile key is required, e.g. event
-// forwarding. It is a no-op if the key is undefined.
-func (b *AcceptedSetBuilder) WithPrimaryMobileKey(key config.MobileKey) *AcceptedSetBuilder {
-	if key.Defined() {
-		b.addMobileKey(key, nil)
-		b.set.primaryMobileKey = key
+// WithMobileKey adds a mobile key. It is a no-op if the value is undefined or already present.
+func (b *AcceptedSetBuilder) WithMobileKey(p MobileKeyParams) *AcceptedSetBuilder {
+	if !p.Value.Defined() || b.set.hasMobileKey(p.Value) {
+		return b
 	}
+	b.set.mobileKeys[p.Value] = acceptedKeyInfo{key: p.Key, expiry: p.Expiry}
 	return b
 }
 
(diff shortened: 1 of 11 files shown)


internal/credential/accepted_set_builder.go
```diff
@@ -16,77 +16,69 @@ type AcceptedSetBuilder struct {
...
-// WithAnchor adds key (if not already present) and designates it as the anchor — the SDK key
-// that owns the environment's upstream connection. It is a no-op if the key is undefined.
-func (b *AcceptedSetBuilder) WithAnchor(key config.SDKKey) *AcceptedSetBuilder {
-	if key.Defined() {
-		b.addSDKKey(key, nil)
-		b.set.anchor = key
+// WithSDKKey adds a server-side SDK key. It is a no-op if the value is undefined or already present
+// (the first metadata recorded for a value wins).
+func (b *AcceptedSetBuilder) WithSDKKey(p SDKKeyParams) *AcceptedSetBuilder {
+	if !p.Value.Defined() || b.set.hasSDKKey(p.Value) {
```
keelerm84, line 43: Question. Previously I think these were arrays, and so checking that we didn't have a duplicate made sense. But now that they are maps, should we allow them to be set again? The value carries metadata like expiry, so perhaps it is worth it?

internal/credential/accepted_set_builder.go
```diff
@@ -16,77 +16,69 @@ type AcceptedSetBuilder struct {
...
+		return b
 	}
-	b.set.sdkKeys[key] = expiry
-}
-
-// WithMobileKey adds a permanent (non-expiring) mobile key. It is a no-op if the key is undefined or
-// already present.
-func (b *AcceptedSetBuilder) WithMobileKey(key config.MobileKey) *AcceptedSetBuilder {
-	b.addMobileKey(key, nil)
+	b.set.sdkKeys[p.Value] = acceptedKeyInfo{key: p.Key, expiry: nil}
```
keelerm84, line 58: Looks like we are doing that here already.

internal/credential/accepted_set_builder.go
```diff
@@ -16,77 +16,69 @@ type AcceptedSetBuilder struct {
...
-// WithPrimaryMobileKey adds key (if not already present) and designates it as the primary mobile
-// key — the singular default (the wire's mobKey) used where one mobile key is required, e.g. event
-// forwarding. It is a no-op if the key is undefined.
-func (b *AcceptedSetBuilder) WithPrimaryMobileKey(key config.MobileKey) *AcceptedSetBuilder {
-	if key.Defined() {
-		b.addMobileKey(key, nil)
-		b.set.primaryMobileKey = key
+// WithMobileKey adds a mobile key. It is a no-op if the value is undefined or already present.
+func (b *AcceptedSetBuilder) WithMobileKey(p MobileKeyParams) *AcceptedSetBuilder {
+	if !p.Value.Defined() || b.set.hasMobileKey(p.Value) {
```
keelerm84, line 65: Pattern repeats here but not below.

internal/credential/rotator.go
```diff
@@ -258,14 +259,21 @@ func reconcileAcceptedKeys[K reconcilableKey](
...
 			continue // already expired; treat as absent
 		}
-		desired[key] = expiry
+		desired[key] = info.expiry
 	}
 	desired[anchor] = nil
 	reconcileAcceptedKeys(desired, r.acceptedSDKKeys, &r.additions, &r.expirations, r.loggers, "SDK key")
+	// Refresh the wire "key" identifier for every key now in the accepted set, including clearing it
+	// when the new payload carries none — otherwise a stale identifier would linger in /status.
+	for key, info := range set.sdkKeys {
```
keelerm84: Iterating through this a second time is not ideal.

Right now, `reconcileAcceptedKeys` takes in a desired map of `[keys]*time.Time`. That function iterates the map and updates the expiration on the accepted keys.

If you updated the `desired` map to array the full value as I point out above, and update this function, then I think you can update both in a single iteration.

internal/credential/rotator.go
```diff
@@ -258,14 +259,21 @@ func reconcileAcceptedKeys[K reconcilableKey](
 // payload may carry for it. The caller must hold the write lock.
 func (r *Rotator) reconcileSDKKeys(set AcceptedSet, anchor config.SDKKey, now time.Time) {
 	desired := make(map[config.SDKKey]*time.Time, len(set.sdkKeys))
-	for key, expiry := range set.sdkKeys {
-		if expiry != nil && !now.Before(*expiry) {
+	for key, info := range set.sdkKeys {
```
keelerm84, line 263: `sdkKeys` is already a map of `[config.SDKKey]acceptedKeyInfo`. Let's filter but keep the full `acceptedKeyInfo`. This will be helpful as I point out below.

internal/credential/rotator.go
```diff
@@ -274,16 +282,23 @@ func (r *Rotator) reconcileSDKKeys(set AcceptedSet, anchor config.SDKKey, now ti
 // and permanent; an empty value means the set declared no mobile key. The caller must hold the lock.
 func (r *Rotator) reconcileMobileKeys(set AcceptedSet, now time.Time) {
 	desired := make(map[config.MobileKey]*time.Time, len(set.mobileKeys))
-	for key, expiry := range set.mobileKeys {
-		if expiry != nil && !now.Before(*expiry) {
+	for key, info := range set.mobileKeys {
```
keelerm84, line 279: Same stuff here.

internal/envfactory/reconcile_helper.go
```diff
@@ -50,34 +44,32 @@ func BuildAcceptedSet(params EnvironmentParams) (credential.AcceptedSet, config.
 		}
 		if k.Value == anchor {
 			anchorInArray = true
-		}
-		if k.Expiry.IsZero() {
-			b.WithSDKKey(k.Value)
+			b.WithAnchor(credential.SDKKeyParams{Value: k.Value, Key: util.PtrOrNil(k.Key)})
```
keelerm84, line 47: Don't bother doing it in this PR, but I wonder about the utility of having the keys being string pointers.

internal/credential/rotator.go
aaron-zeisler: Yep, that makes sense 👍

internal/envfactory/reconcile_helper.go
aaron-zeisler: Ah, this is my mistake. I thought that the key could be optional/empty, and that's why I made it a pointer. But just now I did some additional research and it seems that the key and the value should never be empty. I'll definitely update this. I can make that update as a follow-up PR.

keelerm84: I thought the key was the human-readable identifier? In which case, I think it definitely can be optional. A previously downloaded offline file, or example, or manual configuration.

My point was just pointer vs value type.

aaron-zeisler: Ah ok, the manual configuration makes sense.

Yes, the pointer is there to signify that the value is optional, and to drive home the difference between "empty" and "has some value" (although maybe "empty" and "empty string" are really the same anyways in this case).

internal/credential/accepted_set_builder.go
aaron-zeisler: Talked about this offline, and we'll keep it as-is for now. The code is protecting against having an expiration on either the anchor or the primary mobile key.

internal/credential/rotator.go
```diff
@@ -257,14 +262,18 @@ func reconcileAcceptedKeys[K reconcilableKey](
...
 		}
-		desired[key] = expiry
+		desired[key] = info
 	}
-	desired[anchor] = nil
+	// The anchor is always accepted and permanent regardless of any expiry the payload carries; keep
+	// its wire identifier from the set.
+	anchorInfo := set.sdkKeys[anchor]
+	anchorInfo.expiry = nil
+	desired[anchor] = anchorInfo
```
aaron-zeisler: With the above 3 lines of logic, are the "if set.hasMobileKey() {" checks in the AcceptedSetBuilder.With functions necessary? And if the explicit exclusion of the expiry in BuildAcceptedSet() necessary?

internal/credential/rotator.go
```diff
@@ -28,11 +29,11 @@ type Rotator struct {
 
 	// acceptedSDKKeys is the full set of accepted SDK keys with optional per-key expiry.
 	// A nil expiry means the key is permanent. The anchor is always present with a nil expiry.
-	acceptedSDKKeys map[config.SDKKey]*acceptedKeyInfo
+	acceptedSDKKeys map[config.SDKKey]acceptedKeyInfo
 
 	// acceptedMobileKeys is the full set of accepted mobile keys with optional per-key expiry.
 	// A nil expiry means the key is permanent.
-	acceptedMobileKeys map[config.MobileKey]*acceptedKeyInfo
+	acceptedMobileKeys map[config.MobileKey]acceptedKeyInfo
```
aaron-zeisler, line 36: Note for reviewers: Here I decided to go with values instead of pointers. This caused a handful of random-looking changes throughout the pull request.

keelerm84 (approved):

PR #730, opened 2026-06-29
feat(credential): expose the full accepted key set via AcceptedKeys

aaron-zeisler:
> [!NOTE]
> **Stacked PR — splitting #724:**
> 1. ~~#728 — reject primary mobile key absent from `mobileKeys[]`~~ ✅ merged
> 2. ~~#729 — carry wire key identifiers through the accepted set~~ ✅ merged
> 3. **➡ #730 (this PR)** — expose the full accepted key set via `AcceptedKeys`
> 4. #731 — surface `sdkKeys[]` / `mobileKeys[]` on `/status`
>
> Base: `feat/concurrent-keys`. #731 stacks on this.

## Summary

Expose the full accepted credential set via `Rotator.AcceptedKeys()` and `EnvContext.GetAcceptedKeys()` — every server-side SDK key and mobile key, including the anchor and primary mobile key — so the status endpoint can surface it.

Jira: [SDK-2534](https://launchdarkly.atlassian.net/browse/SDK-2534)

## Background

Second-from-last PR of the stack splitting #724; consumed by #731, which maps the result onto the `sdkKeys[]` / `mobileKeys[]` response arrays.

## Changes

- Promote the per-key metadata struct to an exported **`AcceptedKey{Key, Expiry}`** — a single type used for both the rotator's internal storage and the public snapshot (the credential value is the map key wherever it's stored, so the struct carries no value field).
- Add **`AcceptedKeySet`**: the `Server` / `Mobile` maps keyed by credential value, plus the designated `Anchor` and `PrimaryMobile`. `Rotator.AcceptedKeys()` snapshots both maps and the designations under a single read lock, so callers get a mutually consistent view. `EnvContext.GetAcceptedKeys()` delegates to it.
- Drop the stale `DeprecatedCredentials` doc reference to populating `expiringSdkKey` (that moves to the accepted-set path in #731).

### DIFF
--- a/internal/credential/rotator.go
+++ b/internal/credential/rotator.go
@@ -1,6 +1,7 @@
 package credential
 
 import (
+	"maps"
 	"sync"
 	"time"
 
@@ -8,10 +9,29 @@ import (
 	"github.com/launchdarkly/ld-relay/v8/config"
 )
 
-// acceptedKeyInfo holds per-key metadata for the accepted-set maps.
-type acceptedKeyInfo struct {
-	expiry *time.Time // nil = permanent
-	key    *string    // wire "key" identifier — non-secret human-readable name; nil when absent
+// AcceptedKey is the metadata for one accepted credential: its optional expiry and optional wire
+// "key" identifier. The credential value itself is the map key wherever AcceptedKey is stored — the
+// rotator's accepted-key maps, the builder's AcceptedSet, and the AcceptedKeySet returned by
+// AcceptedKeys.
+type AcceptedKey struct {
+	// Expiry is the key's expiry. A nil expiry means the key is permanent.
+	Expiry *time.Time
+	// Key is the non-secret wire "key" identifier — a human-readable name. Nil when the source carried
+	// none (manual configuration, or an old-format payload predating concurrent keys).
+	Key *string
+}
+
+// AcceptedKeySet is a point-in-time snapshot of an environment's full accepted credential set,
+// returned by Rotator.AcceptedKeys. Server and Mobile are keyed by credential value (the secret);
+// the value AcceptedKey carries that key's metadata. Anchor and PrimaryMobile name the designated
+// keys within Server and Mobile. The status endpoint maps Server/Mobile to the sdkKeys[]/mobileKeys[]
+// arrays and uses Anchor to mark the anchor entry. Reads of the maps and the designations are taken
+// under a single lock, so they are mutually consistent.
+type AcceptedKeySet struct {
+	Server        map[config.SDKKey]AcceptedKey
+	Mobile        map[config.MobileKey]AcceptedKey
+	Anchor        config.SDKKey
+	PrimaryMobile config.MobileKey
 }
 
 type Rotator struct {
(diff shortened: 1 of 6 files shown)


internal/credential/rotator.go
```diff
@@ -314,3 +338,25 @@ func (r *Rotator) reconcileEnvironmentID(set AcceptedSet) {
 	r.primaryEnvironmentID = set.envID
 	r.additions = append(r.additions, set.envID)
 }
+
+// acceptedKeysOf snapshots one accepted-key map (SDK or mobile) into a slice of AcceptedKey tagged
+// with kind. Order is unspecified — the status endpoint does not depend on it. The caller must hold
+// at least the read lock.
```
keelerm84: This function isn't defined on any type, so there isn't a read lock to hold. I get what it's suggesting, but it seems like this would be better suited to the call site rather than the function definition.

internal/credential/rotator.go
```diff
@@ -314,3 +338,25 @@ func (r *Rotator) reconcileEnvironmentID(set AcceptedSet) {
...
+
+// AcceptedKeys returns every accepted credential — all server-side SDK keys and all mobile keys,
+// including the anchor and the primary mobile key. The status endpoint partitions the result by Type
+// to populate the full sdkKeys[] / mobileKeys[] arrays. Order is unspecified.
+func (r *Rotator) AcceptedKeys() []AcceptedKey {
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+
+	keys := acceptedKeysOf(r.acceptedSDKKeys, KeyTypeServer)
+	return append(keys, acceptedKeysOf(r.acceptedMobileKeys, KeyTypeMobile)...)
```
keelerm84: Instead of having to walk the keys to tag with their types, can we just return a simpler structure that is 

```go
struct {
    server: [cred]acceptedKeyInfo
    mobile: [cred]acceptedKeyInfo
}
```

internal/credential/rotator.go
aaron-zeisler: Ah, thanks for pointing this out. In a previous iteration, this function included a lot more logic (that I killed because it was slop). Now that it's been slimmed down, I think it could be inline, The only caller is the function below.

UPDATE: I like your suggestion below - that kills all the code in this function 👍

internal/credential/rotator.go
aaron-zeisler: Nice

aaron-zeisler: The function is consumed by other packages, so I may still need to define an exported type ... I'll see if I can use the existing types first, but it may need to be
```go
func (r *Rotator) AcceptedKeys() map[KeyType]AcceptedKey
```
where AcceptedKey no longer contains the KeyType within it. Or maybe it's best to just export acceptedKeyInfo.

(Some of these notes are for my agent 🙂 )

internal/credential/rotator.go
```diff
@@ -1,17 +1,37 @@
...
 	"time"
 
 	"github.com/launchdarkly/go-sdk-common/v3/ldlog"
 	"github.com/launchdarkly/ld-relay/v8/config"
 )
 
-// acceptedKeyInfo holds per-key metadata for the accepted-set maps.
-type acceptedKeyInfo struct {
-	expiry *time.Time // nil = permanent
-	key    *string    // wire "key" identifier — non-secret human-readable name; nil when absent
```
aaron-zeisler, line 14: Note for reviewers: I decided to combine the internal data structure (acceptedKeyInfo) with the external-facing response object (AcceptedKey).

keelerm84 (approved):

PR #741, opened 2026-07-08
feat: Re-anchor big-segment sync onto the new anchor

keelerm84:
## Summary

When an environment re-anchors (its upstream SDK key changes), the big-segment synchronizer must follow the new anchor. The synchronizer bakes its SDK key in at construction and authenticates its poll/stream requests with that key directly, and it is not restartable (single-use `startOnce`/`closeOnce`). Left unchanged it would keep polling with the retired key after a re-anchor.

This recreates the synchronizer on the new anchor as part of the synchronous re-anchor commit:

- A `makeBigSegmentSync(anchor)` closure captured at construction binds the transport-only httpConfig, store, URIs, env ID, and loggers; only the anchor SDK key varies per rebuild.
- `commitReanchor` calls `reanchorBigSegmentSync(newAnchor)` under `c.mu`: build the replacement on the new key, re-attach its update consumer, start it iff a big segment had already appeared (so synchronization continues without a gap), then close the old synchronizer (which ends its consumer goroutine). No-op when big segments are not configured, and never runs on a rolled-back re-anchor.
- The update consumer was extracted to `consumeBigSegmentUpdates`, spawned per synchronizer instance so each rebuild gets its own consumer bound to its own channel.
- The store-update sink's read of the (now runtime-mutable) synchronizer field is synchronized, and `setBigSegmentsExist` starts the current synchronizer under the lock, so a concurrent re-anchor can never leave the wrong instance started.

Stacked on #739; targets `feat/concurrent-keys`.

---

> [!NOTE]
> **Medium Risk**
> Touches credential re-anchor commit path and adds goroutine lifecycle around synchronizer swaps; mitigated by extensive tests including rollback, ordering, and race regression coverage.
> 
> **Overview**
> **Big-segment sync now follows SDK re-anchors** instead of staying on the retired anchor key. On `commitReanchor`, `reanchorBigSegmentSync` rebuilds the synchronizer (key is fixed at construction), re-attaches an update consumer, **starts** the replacement immediately if big segments were already active, then **closes** the old instance.
> 
> Construction stores a **`makeBigSegmentSync(anchor)`** closure so rebuilds reuse transport/store/URIs and only swap the anchor SDK key. Update handling moves to **`consumeBigSegmentUpdates`**, with one consumer goroutine per synchronizer instance.
> 
> Concurrency fixes: **`setBigSegmentsExist`** starts the current synchronizer under `c.mu` so a concurrent re-anchor cannot start a retired instance; store-update paths use **`bigSegmentSyncConfigured()`** (RLock) because `bigSegmentSync` is reassigned during re-anchor.
> 
> Tests cover started-sync continuity, rollback (no rewire), not-configured no-op, client-side invalidation after re-anchor, ordering before first segment, sequential A→B→C re-anchors, and concurrent store updates; mocks now close the update channel on `Close`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d9a73e3eb07cab8998a2f08f62113b4fbcf3ac14. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/relayenv/env_context_impl.go
+++ b/internal/relayenv/env_context_impl.go
@@ -89,19 +89,23 @@ type EnvContextImplParams struct {
 }
 
 type envContextImpl struct {
-	mu                        sync.RWMutex
-	clients                   map[config.SDKKey]sdks.LDClientContext
-	storeAdapter              *store.SSERelayDataStoreAdapter
-	loggers                   ldlog.Loggers
-	identifiers               EnvIdentifiers
-	secureMode                bool
-	envStreams                *streams.EnvStreams
-	streamProviders           []streams.StreamProvider
-	handlers                  map[streams.StreamProvider]map[credential.SDKCredential]http.Handler
-	jsContext                 JSClientContext
-	evaluator                 ldeval.Evaluator
-	eventDispatcher           *events.EventDispatcher
-	bigSegmentSync            bigsegments.BigSegmentSynchronizer
+	mu              sync.RWMutex
+	clients         map[config.SDKKey]sdks.LDClientContext
+	storeAdapter    *store.SSERelayDataStoreAdapter
+	loggers         ldlog.Loggers
+	identifiers     EnvIdentifiers
+	secureMode      bool
+	envStreams      *streams.EnvStreams
+	streamProviders []streams.StreamProvider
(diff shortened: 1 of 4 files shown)


aaron-zeisler (approved):
Excellent 👍

PR #742, opened 2026-07-08
feat: Build stream handlers on demand instead of per credential

keelerm84:
## Summary

Relay built one HTTP stream handler per `(credential, filter, stream provider)` and stored the full fan-out on each env context (`envContextImpl.handlers`). Every handler in a `(filter, provider)` slot is byte-identical except for the credential-derived channel id, so the stored fan-out was pure duplication -- `C x F x P` handlers per env (~2,000 at Block scale: 50 keys x 10 filters x 4 providers).

`GetStreamHandler` now builds the handler on demand from the request's already-authenticated credential, scoping it with the env's (immutable) filter key:

```go
if h := streamProvider.Handler(sdkauth.NewScoped(c.filterKey, cred)); h != nil {
    return h
}
return http.HandlerFunc(invalidStreamHandler)
```

The per-credential handler map (field, init, both build loops, and the per-credential teardown in `removeCredential`) is removed. `envStreams.AddCredential` (which registers the eventsource channel Repository) and the connection mapping stay per credential -- those are the real per-credential state and the authoritative routing/authorization gate.

This is behavior- and authorization-equivalent to the prior code: a request only reaches `GetStreamHandler` after the middleware resolves the env via the connection-mapping allow-list (scoped by filter + credential), so the credential always belongs to this env and the on-demand build is byte-for-byte the same channel the prebuilt handler would have served. Single-key behavior is unchanged; multi-key routing is unchanged.

Result: zero stored stream handlers per env instead of `C x F x P`.

Stacked on #741 (T2.d); base `mk/sdk-2543/big-segment-sync-from-anchor` until that merges.

---

> [!NOTE]
> **Medium Risk**
> Touches core streaming request path and credential lifecycle wiring; behavior is intended to be equivalent but relies on middleware having already authorized the credential before on-demand handler construction.
> 
> **Overview**
> **Stream handlers are no longer pre-built and cached per credential.** `GetStreamHandler` now calls `streamProvider.Handler` with a scoped `(filterKey, credential)` on each request and returns that handler (or the existing 404 invalid-stream handler when the provider returns nil). The `envContextImpl.handlers` map and all logic that populated or pruned it at env init, credential add, and credential remove are gone.
> 
> **Per-credential routing state is unchanged:** `envStreams.AddCredential` / `RemoveCredential` and connection mapping still run when credentials are registered or revoked; only the redundant handler fan-out was removed. `GetStreamHandler` no longer takes `c.mu` because `filterKey` is fixed after construction.
> 
> New unit tests in `env_context_handler_fanout_test.go` cover on-demand scoping, independent resolution per credential, and wrong-kind credentials serving 404. An existing reconcile test comment was updated to match the slimmer credential registration path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 36d40e8bc599c94fe2333a220b0f6a497f83d1b0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/relayenv/env_context_impl.go
+++ b/internal/relayenv/env_context_impl.go
@@ -898,12 +898,12 @@ func (c *envContextImpl) GetLoggers() ldlog.Loggers {
 	return c.loggers
 }
 
-func (c *envContextImpl) GetStreamHandler(streamProvider streams.StreamProvider, cred credential.SDKCredential) http.Handler {
+func (c *envContextImpl) GetStreamHandler(streamProvider streams.StreamProvider, credential credential.SDKCredential) http.Handler {
 	// Build the handler on demand rather than storing one per (credential, provider): every handler in a
 	// (filter, provider) slot is identical except for the credential-derived channel id, which we resolve
 	// here from the request's already-authenticated credential. c.filterKey is immutable after
 	// construction, so this needs no lock.
-	if h := streamProvider.Handler(sdkauth.NewScoped(c.filterKey, cred)); h != nil {
+	if h := streamProvider.Handler(sdkauth.NewScoped(c.filterKey, credential)); h != nil {
 		return h
 	}
 	return http.HandlerFunc(invalidStreamHandler)
--- a/internal/relayenv/env_context_impl.go
+++ b/internal/relayenv/env_context_impl.go
@@ -89,19 +89,23 @@ type EnvContextImplParams struct {
 }
 
 type envContextImpl struct {
-	mu                        sync.RWMutex
-	clients                   map[config.SDKKey]sdks.LDClientContext
-	storeAdapter              *store.SSERelayDataStoreAdapter
-	loggers                   ldlog.Loggers
-	identifiers               EnvIdentifiers
-	secureMode                bool
-	envStreams                *streams.EnvStreams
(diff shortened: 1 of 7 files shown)


internal/relayenv/env_context_impl.go
```diff
@@ -918,14 +898,15 @@ func (c *envContextImpl) GetLoggers() ldlog.Loggers {
 	return c.loggers
 }
 
-func (c *envContextImpl) GetStreamHandler(streamProvider streams.StreamProvider, credential credential.SDKCredential) http.Handler {
-	c.mu.RLock()
-	defer c.mu.RUnlock()
-	h := c.handlers[streamProvider][credential]
-	if h == nil {
-		return http.HandlerFunc(invalidStreamHandler)
+func (c *envContextImpl) GetStreamHandler(streamProvider streams.StreamProvider, cred credential.SDKCredential) http.Handler {
```
aaron-zeisler: Nit: is the 'credential' -> 'cred' variable name change necessary here in the function signature?

aaron-zeisler (approved):
👍 really great change

PR #746, opened 2026-07-13
feat: concurrency limit for server-side streaming endpoints

JackDanger:
Adds an optional concurrency limit for `/all` and `/flags`. Requests over the limit get `503` + `Retry-After` before reaching the handler. If unset, a default is derived from detected available memory (cgroup limit, else total system memory) so it adapts across instance/pod sizes without config changes. Also adds a `stream_admissions_rejected` metric next to the existing connections gauge.


kinyoklion:
Closing this in favor of an init concurrency limit.

PR #756, opened 2026-07-16
test(concurrent-keys): even more integration tests

aaron-zeisler:
## Summary

Two more integration tests from [SDK-2609](https://launchdarkly.atlassian.net/browse/SDK-2609). These required a bit more setup than those in #755. Both are test-only — no production code changes.

## Changes

- **Offline anchor rotation via archive reload**: reloading the offline config with a brand-new anchor key swaps it in — the old anchor stops while the non-anchor key and its open stream are undisturbed.
- **Cache reload survives a restart** (`redis_unit_tests`): a multi-key environment persisted to the real Redis-backed auto-config cache is reloaded, with its `sdkKeys[]`/`mobileKeys[]` arrays intact, by a fresh `StreamManager` whose config stream delivers nothing — so the cache is the only possible source. Exercises the production `StreamManager` + `redisStore` against a real Redis.

[SDK-2609]: https://launchdarkly.atlassian.net/browse/SDK-2609?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Low Risk**
> Test-only changes (Redis-tagged integration test and offline relay harness); no runtime behavior modified in this diff.
> 
> **Overview**
> Adds two **SDK-2609** regression tests; no production code in this diff.
> 
> **Redis AutoConfig cache (restart survival):** New `TestConcurrentKeysCacheReloadSurvivesRestart` (`redis_unit_tests`) seeds Redis via a live `StreamManager`, shuts it down, then starts a second manager on the same cache with a silent config stream. It asserts the environment is reloaded from cache with full `AcceptedSDKKeys` / `AcceptedMobileKeys` (anchor + extra for both SDK and mobile).
> 
> **Offline multi-key anchor rotation:** Adds `multiKeyArchiveEnvWithAnchor` and `TestConcurrentKeysOffline_AnchorRotationViaArchiveReload`, which reloads file-data with a new anchor while a downstream SDK stays connected on a non-anchor key. Expects no new upstream client, stream stays open, new anchor + extra key auth, old anchor rejected.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5d1818d2846d3a7b6a6c18d54b7c6d0b14b77621. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/autoconfigcache/reload_restart_redis_test.go
+++ b/internal/autoconfigcache/reload_restart_redis_test.go
@@ -1,0 +1,161 @@
+//go:build redis_unit_tests
+// +build redis_unit_tests
+
+package autoconfigcache
+
+// Verifies that a multi-key environment written to the real Redis-backed AutoConfig cache survives a
+// process restart: a fresh StreamManager, given the same Redis cache but a config stream that delivers
+// nothing, reloads the environment from the cache with its sdkKeys[]/mobileKeys[] arrays intact.
+//
+// This is the restart-survival half of the "cache integrity across restart" scenario (SDK-2609 #10);
+// the malformed-put-preserves-cache half is covered by the StreamManager unit test
+// TestMalformedCredentialPayloadPreservesEnvironmentCache. It exercises the production classes
+// (StreamManager + redisStore) against an actual Redis, with no test doubles for the cache itself.
+//
+// Requires a Redis server on localhost (the redis_unit_tests build tag; CI provides one).
+
+import (
+	"context"
+	"net/url"
+	"testing"
+	"time"
+
+	"github.com/launchdarkly/ld-relay/v8/config"
+	"github.com/launchdarkly/ld-relay/v8/internal/autoconfig"
+	"github.com/launchdarkly/ld-relay/v8/internal/envfactory"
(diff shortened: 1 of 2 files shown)


relay/relay.go
```diff
@@ -82,6 +82,12 @@ type relayInternalOptions struct {
 	loggers               ldlog.Loggers
 	clientFactory         sdks.ClientFactoryFunc
 	archiveManagerFactory func(path string, monitoringInterval time.Duration, environmentUpdates filedata.UpdateHandler, loggers ldlog.Loggers) (filedata.ArchiveManagerInterface, error)
+	// autoConfigCache, when non-nil, overrides the persistent AutoConfig cache that would otherwise be
```
keelerm84: Generally, I don't like modifying something in production to support testing. We already have some unit tests that run against an actual redis instance. Could we do the same tests using that?

keelerm84 (approved):

PR #757, opened 2026-07-17
fix: Encode FDv2 stream events in a single jwriter pass

keelerm84:
The FDv2 event builders (MakeEventsForUpToDate, MakeEventsForSetBasis,
MakeEventsForApplyDelta) previously went through
ldservicesv2.StreamingProtocol, whose pushEvent calls json.Marshal on each
event. For put-object events this re-parses and compacts the item JSON that
was already produced by jwriter on the replay path, encoding the entire
payload twice. It also made a test-helpers package part of the production
hot path, with a panic on marshal errors.

The builders now render each event's data directly with jwriter, embedding
the pre-serialized object bytes verbatim. Wire output is unchanged aside
from encoding/json's optional HTML escaping (matching the existing FDv1
event encoders, which have never escaped).

BenchmarkMakeEventsForSetBasisWithManyFlags (new, mirrors the full-transfer
replay path): 744us -> 368us per op (-50%).

---

> [!NOTE]
> **Medium Risk**
> Touches the FDv2 SSE hot path for all connected SDK streams; behavior is meant to be wire-compatible but any encoding mismatch could affect clients.
> 
> **Overview**
> FDv2 stream event builders (`MakeEventsForUpToDate`, `MakeEventsForSetBasis`, `MakeEventsForApplyDelta`) no longer use `ldservicesv2.StreamingProtocol`, which **json.Marshal**’d each event and re-encoded put-object payloads that were already serialized on the replay path.
> 
> Each FDv2 event is now built with **jwriter** into a new **`fdv2Event`** type (data fixed at construction). Put events embed **`change.Object`** via **`Raw`**, so item JSON is not parsed and compacted again. The test-helpers streaming builder is removed from the production path.
> 
> Wire shape is intended to match the previous output; encoding may differ only where **`encoding/json`** would HTML-escape (aligned with existing FDv1 jwriter encoders). A benchmark **`BenchmarkMakeEventsForSetBasisWithManyFlags`** exercises the full-transfer replay-style path (~50% faster per the PR description).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0343e13ed643d356dfd607f28c0aedb1455032f8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


aaron-zeisler (approved):
👍 looks good

Not blocking: There's a great benchmark test, but is there a test that verifies the encoded output is the same before-vs-after this change? (that test might already exist, but didn't need to change in this PR).

kinyoklion (approved):

PR #759, opened 2026-07-20
fix(store): prevent a startup race from leaving the environment with a closed data store

aaron-zeisler:
## Summary

Fixes a startup race in the store adapter that could leave an environment reading from a data store that had already been closed. Found during the final pre-merge review of concurrent keys ([SDK-2702](https://launchdarkly.atlassian.net/browse/SDK-2702)).

## What goes wrong

When an environment starts up, two things can try to set up its data store at the same time: the initial client's build, and the first key-rotation re-anchor's build. Neither holds the environment lock, so both can look, see no store yet, and each build its own.

Whichever finished last used to win — it overwrote the store pointer unconditionally. If that "winner" was the client we then throw away (because it was superseded), closing it tore down the very store the environment was serving from. The result: evaluations and stream queries read a closed store, while the live client quietly feeds a store nothing is reading — and it stays that way until the next re-anchor rebuilds everything.

The race detector can't see it (the locking was technically correct), and the existing handover race test only starts after the first store is installed, so it never exercised this window.

## The fix

`Build` now holds the adapter lock across the whole build, so two concurrent builds can't each construct and install their own store. The second caller blocks until the first has installed its store, then adopts it via the normal handover path — one build at a time, so the last-writer-wins hazard simply can't arise.

Holding the lock across the build blocks `GetStore()` only while a store is being built from nothing, which is the first build at startup — every later build, including a re-anchor handover, returns the already-built wrapper without calling the wrapped factory. And for the persistent stores relay supports (Redis, DynamoDB, Consul) the wrapped factory's `Build` only allocates a client or pool; it doesn't connect. The first datastore round-trip happens later on `Init`/`Get`, outside the lock. So the lock is never held across datastore I/O.

A regression test (`store_concurrent_build_test.go`) forces the interleaving deterministically with a gated factory: while the first build stalls inside the factory, a second build blocks on the lock; once released, both return the same wrapper and the factory is built exactly once.

## Alternatives considered

- **Re-check after building and adopt-and-discard** — let both builds run, then have the late one notice the wrapper the other installed, adopt it, and close its own. This never holds the lock across the build, but it adds a subtler "last writer wins is a hazard" path and more machinery to guard a window that (since the factory doesn't connect) is only microseconds wide. Rejected in favor of the simpler invariant.
- **A separate mutex serializing only `Build`** — prevents the double-build without blocking `GetStore()` readers at all, but adds a second lock and more state than just holding the existing one across a build that does no I/O. Rejected as unnecessary.

[SDK-2702]: https://launchdarkly.atlassian.net/browse/SDK-2702

---

> [!NOTE]
> **Medium Risk**
> Changes locking around environment data-store creation at startup/re-anchor; incorrect serialization could block builds or affect handover, but scope is limited to the adapter and covered by a targeted regression test.
> 
> **Overview**
> Fixes a startup race where the initial anchor client’s `Build` and a concurrent re-anchor `Build` could each create a wrapper; the last install could win, and closing the superseded client would tear down the store the environment still served.
> 
> **`SSERelayDataStoreAdapter.Build`** now holds the adapter mutex for the full build (including the wrapped factory), so concurrent builds serialize: the first installs one wrapper and the second reuses it via the existing handover path instead of building again.
> 
> Adds **`TestConcurrentBuildSerializesAndSharesOneWrapper`** with a gated factory to prove the second build blocks until the first finishes, both share one wrapper, the underlying factory runs once, and closing the superseded holder does not close the live store.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4a899ff6aba68465e7736012fe597f37cbb6019c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/store/relay_feature_store.go
+++ b/internal/store/relay_feature_store.go
@@ -78,14 +78,21 @@ func NewSSERelayDataStoreAdapter(
 func (a *SSERelayDataStoreAdapter) Build(
 	context subsystems.ClientContext,
 ) (subsystems.DataStore, error) {
+	// The lock is held across the whole build so two concurrent Build calls cannot each construct and
+	// install their own wrapper (the last writer would win, and if its client were later discarded as
+	// superseded, its Close would tear down the store the adapter is serving). A second caller blocks
+	// until the first installs its store, then adopts it via the fast path below. This can block
+	// GetStore only while a store is built from nothing — the first build at startup — since every
+	// later Build, including a re-anchor handover, returns the already-built wrapper without calling
+	// the wrapped factory.
 	a.mu.Lock()
+	defer a.mu.Unlock()
+
 	if existing := a.store; existing != nil {
 		if sw, ok := existing.(*streamUpdatesStoreWrapper); ok && sw.acquire() {
-			a.mu.Unlock()
 			return sw, nil
 		}
 	}
-	a.mu.Unlock()
 
 	wrappedStore, err := a.wrappedFactory.Build(context)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


internal/store/relay_feature_store.go
```diff
@@ -98,8 +98,22 @@ func (a *SSERelayDataStoreAdapter) Build(
 	)
 
 	a.mu.Lock()
-	defer a.mu.Unlock()
```
keelerm84, line 101: Instead of re-checking, why aren't we just holding the lock longer?

keelerm84 (changes requested):

internal/store/relay_feature_store.go
aaron-zeisler: Yeah, this is one of the alternative solutions that was floated.

I assumed that the `a.wrappedFactory.Build()` call above would be slow because it might reach out to the data source (Redis or DynamoDB, etc...). But I just dug into it further, and I see that those Build() functions either don't reach out to the actual systems, or they do so lazily. So the call to `wrappedFactory.Build()` is actually pretty cheap.

I'll go with your idea and just hold the lock through this entire function 👍

keelerm84 (approved):

PR #761, opened 2026-07-20
fix(concurrent-keys): reject mobile-key payloads that arrive without a primary mobile key

aaron-zeisler:
## Summary

When trusted-source credentials arrive, `BuildAcceptedSet` validates them. On the SDK-key side it already rejects a payload whose anchor key is missing, or is defined but not listed in `sdkKeys[]`. The mobile side was missing the matching check: a payload that carried mobile keys (`mobileKeys[]` non-empty) but named no primary (`mobKey` undefined) was accepted.

That's a problem because the reconcile then clears the environment's primary mobile key without repointing event forwarding — so mobile analytics events keep going out under the old, possibly revoked, primary. The result is silent misattribution instead of a loud rejection.

This adds the symmetric guard: when there are mobile keys but no designated primary, `BuildAcceptedSet` rejects the payload as malformed (`NewPrimaryMobileKeyMissingError`), so it rides the existing malformed-payload path — previous credentials preserved, stream restarted. The doc comment enumerating the malformed cases is updated to cover both mobile cases (the existing "primary defined but not in the array" case and this new "no primary" case).

Two shapes stay valid, and are covered by tests:

- An environment with no mobile keys at all (empty array, undefined `mobKey`) — e.g. server-side-only environments.
- Old-format payloads, which synthesize the mobile array from `mobKey` alone, so an undefined `mobKey` just yields an empty array.

---

> [!NOTE]
> **Medium Risk**
> Changes trusted-source credential validation and RAC malformed-payload handling for mobile keys; behavior is narrow, defensive, and covered by tests.
> 
> **Overview**
> **`BuildAcceptedSet`** now treats a non-empty **`mobileKeys[]`** with an undefined **`mobKey`** as malformed (via **`NewPrimaryMobileKeyMissingError`**), matching the existing SDK anchor rules. Accepting that shape would clear the primary on reconcile while event forwarding could keep using the old primary.
> 
> Malformed payloads still follow the existing path: prior credentials stay in place and the RAC stream restarts. Docs list this as a third mobile malformed case; unit tests cover reject/accept boundaries and extend stream restart tests to include this payload shape alongside undefined anchor SDK keys.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 85f8a02ef031fde3a4bfdec9eeda75eb887d47f5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/autoconfig/stream_manager_errors_test.go
+++ b/internal/autoconfig/stream_manager_errors_test.go
@@ -34,55 +34,81 @@ func eventShouldCauseStreamRestart(t *testing.T, event httphelpers.SSEEvent) {
 }
 
 // A credential payload that is valid JSON and a structurally valid event, but whose credential set
-// cannot be built (here: an undefined anchor SDK key), must be caught at the parse boundary: the
-// previous state is preserved (no AddEnvironment/UpdateEnvironment dispatched) and the stream is
-// restarted so the backend resends a fresh put (design §9). This is verified for both patch and put,
-// since both paths run the validation before the version is recorded.
+// cannot be built (e.g. an undefined anchor SDK key, or mobile keys with no designated primary), must
+// be caught at the parse boundary: the previous state is preserved (no AddEnvironment/UpdateEnvironment
+// dispatched) and the stream is restarted so the backend resends a fresh put (design §9). This is
+// verified for both patch and put, since both paths run the validation before the version is recorded.
 func TestMalformedCredentialPayloadCausesStreamRestart(t *testing.T) {
-	malformedEnv := testEnv1
-	malformedEnv.SDKKey = envfactory.SDKKeyRep{Value: config.SDKKey("")} // undefined anchor
+	// Each shape is a distinct way BuildAcceptedSet rejects a structurally malformed payload; all ride
+	// the same malformed-payload machinery.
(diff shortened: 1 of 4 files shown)


keelerm84:
> The mobile side was missing the matching check: a payload that carried mobile keys (mobileKeys[] non-empty) but named no primary (mobKey undefined) was accepted.

This work is treating this as if it's too late to add this check. None of this is released yet. We can just add this check, and then none of this should matter, right?

aaron-zeisler:
> > The mobile side was missing the matching check: a payload that carried mobile keys (mobileKeys[] non-empty) but named no primary (mobKey undefined) was accepted.
> 
> This work is treating this as if it's too late to add this check. None of this is released yet. We can just add this check, and then none of this should matter, right?

@keelerm84 just to make sure i'm picking up what you're putting down: you mean to guarantee this check in the backend service that emits the RAC payload?

This PR's code is adding the check in our BuildAcceptedSet function. That's the first place in our backend code where logic is applied to the incoming RAC payload. Apologies if this PR's description is both too verbose and not clear enough to describe what's happening 😞

keelerm84 (approved):

PR #762, opened 2026-07-20
refactor(concurrent-keys): remove production-dead DeprecatedCredentials accessors

aaron-zeisler:
## Summary

Removes the production-dead `DeprecatedCredentials` accessor trio and rewrites its test callers to read expiry data from the accepted-set snapshot instead.

- Jira: [SDK-2702](https://launchdarkly.atlassian.net/browse/SDK-2702)
- Stacked on #760 (base branch `aaronz/SDK-2702/dead-code-removal`).

## Background

The trio — `Rotator.DeprecatedCredentials`, `envContextImpl.GetDeprecatedCredentials`, and the `EnvContext.GetDeprecatedCredentials` interface entry — had no remaining production consumers. Their last caller, the `/status` endpoint, now reads `GetAcceptedKeys`, which exposes strictly more: the full accepted set with per-key expiry plus the anchor designation. "Which keys are being phased out" is derivable from that snapshot (non-anchor server keys carrying a future expiry), so the dedicated accessors are pure surface area.

They were deliberately kept during development because the test rewrites looked like they'd cost more than the accessors were worth. This PR is the actual diff, so that call can be made against something concrete rather than a guess.

## Decision requested

These accessors were kept on purpose — the worry was that the test-side rewrite would be messier than just leaving three thin methods in place. Here's the honest accounting now that it's done:

| Metric | Value |
| --- | --- |
| Production lines | +1 / −27 (net **−26**); comment reword is the only addition |
| Production items removed | 3 accessors across 3 files |
| Test lines | +65 / −22 (net **+43**) across 5 files |
| Test call sites rewritten | 19 sites across 5 files |
| Shared helper cost | ~36 of the added test lines are the helper (one ~11-line `deprecatedSDKKeys` per package) |
| Net overall | +66 / −49 (net **+17**) |

A small `deprecatedSDKKeys` helper per test package (`credential`, `relayenv`, `relay`) sources the expiring keys from `AcceptedKeys().Server` / `GetAcceptedKeys()`, filtered to non-anchor server keys with a non-nil expiry — exactly what the removed methods computed. With the helper in place, every call site became a 1:1 swap (`r.DeprecatedCredentials()` → `deprecatedSDKKeys(r)`, `env.GetDeprecatedCredentials()` → `deprecatedSDKKeys(env)`); no inline map-filtering leaked into any assertion, and every behavioral assertion is preserved unchanged.

The one honest wart: the helper is duplicated across three packages because each operates on a different type (`*Rotator` vs `EnvContext`), and sharing it would mean adding non-test production API — which defeats the point of removing production API. So it's three ~11-line test-local functions.

My read: the rewrite came out clean, and the original worry was mostly unfounded once the helper absorbed the filtering. But it is your call. If the test-side cost still outweighs the cleanup, close this PR — no hard feelings.

[SDK-2702]: https://launchdarkly.atlassian.net/browse/SDK-2702?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/credential/rotator.go
+++ b/internal/credential/rotator.go
@@ -133,25 +133,6 @@ func (r *Rotator) allCredentials() []SDKCredential {
 	return creds
 }
 
-// DeprecatedCredentials returns the SDK keys being phased out — every accepted SDK key, other than the
-// anchor, that carries a future expiry. (Per-key expiry is stored as data on the accepted entry; the
-// cleanup ticker drops the key once it elapses.)
-//
-// Mobile keys are deliberately not returned even though they expire the same way SDK keys do — carried
-// as per-key expiry and dropped by the same cleanup ticker.
-func (r *Rotator) DeprecatedCredentials() []SDKCredential {
-	r.mu.RLock()
-	defer r.mu.RUnlock()
-
-	var out []SDKCredential
-	for key, info := range r.acceptedSDKKeys {
-		if info.Expiry != nil && key != r.anchorKey {
-			out = append(out, key)
-		}
-	}
-	return out
-}
-
 // AllCredentials returns every accepted credential: every accepted SDK key, every accepted mobile
 // key (including those carrying a future expiry — they still authenticate until the cleanup ticker
 // drops them), and the environment ID.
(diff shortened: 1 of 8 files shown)


aaron-zeisler:
After removing the unused production code, we still need the functionality. And that functionality is required across three packages, so a helper function is copy-pasted in a few places. I'm closing the PR since I think it's better to keep the existing production code, even if it's only called from tests.

PR #765, opened 2026-07-20
fix(concurrent-keys): defensive hardening bundle from final review

aaron-zeisler:
## Summary

Two small defensive-hardening changes surfaced during the final concurrent-keys review, bundled so each can be judged on its real diff. Every item is a single, independently-droppable commit with no cross-item coupling.

| # | What it does |
|---|--------------|
| 1 | Document the client-factory store-release contract on `buildNewAnchorClient` and `sdks.ClientFactoryFunc` |
| 2 | `GetInitError` takes an `RLock` instead of a full `Lock` (called per-env on `/status`) |

### 1 — Client-factory store-release contract
If a client factory returned `(nil, err)` after the SDK had already built the data store, the store's refcount would never be released, pinning it past env `Close`. The real SDK never returns that shape, but a test factory could. There's no cheap sound guard: `buildNewAnchorClient` can't tell whether the failed call acquired a store ref, so releasing one unconditionally could double-release the still-serving previous anchor's store. This documents the invariant on `buildNewAnchorClient` and `sdks.ClientFactoryFunc` rather than adding a guard.

### 2 — Read lock in GetInitError
`GetInitError` took a full write lock for a pure read that `/status` calls once per environment. It now uses `RLock`.

### DIFF
--- a/internal/relayenv/env_context_impl.go
+++ b/internal/relayenv/env_context_impl.go
@@ -970,8 +970,8 @@ func (c *envContextImpl) GetFilter() config.FilterKey {
 }
 
 func (c *envContextImpl) GetInitError() error {
-	c.mu.Lock()
-	defer c.mu.Unlock()
+	c.mu.RLock()
+	defer c.mu.RUnlock()
 
 	return c.initErr
 }
--- a/internal/relayenv/env_context_impl.go
+++ b/internal/relayenv/env_context_impl.go
@@ -724,6 +724,15 @@ func (c *envContextImpl) reanchor(change *credential.AnchorChange) bool {
 // half-built client and logged a structured error. initErr is deliberately left untouched on failure:
 // it feeds the request middleware, and setting it to the new anchor's ErrInitializationFailed would 401
 // an env that is serving fine on the previous anchor.
+//
+// Factory contract (see sdks.ClientFactoryFunc): a factory whose construction builds the environment's
+// data store must return a non-nil client even on init failure, so that the client.Close() below releases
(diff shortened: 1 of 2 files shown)


internal/envfactory/reconcile_helper.go
```diff
@@ -5,6 +5,14 @@ import (
 	"github.com/launchdarkly/ld-relay/v8/internal/util"
 )
 
+// maxAcceptedKeysPerKind caps the number of entries BuildAcceptedSet will accept in sdkKeys[] or
+// mobileKeys[] before treating the payload as malformed. It is deliberately far larger than any
+// plausible environment — even aggressive key rotation leaves only a handful of keys valid within the
+// grace window at once — so it trips only on a garbage payload, never on legitimate rotation. It is a
+// sanity bound, not a product limit; err high, because rejecting a legitimate payload sends the RAC
+// handler into a reconnect loop against the same oversized put.
+const maxAcceptedKeysPerKind = 10000
```
keelerm84: What is the reason for adding this limit? What are we hoping to guard against?

Also worth noting, upstream already has limits in place for this. I believe the default is 50.

aaron-zeisler: Ah, that's interesting that there may be limits in place on the FD side. This is just guarding against processing a ridiculously huge payload. A safeguard against resource exhaustion (lots of memory consumption). I think the likelihood is low that we'll get giant payloads unless the payload is buggy or if a customer's agent goes crazy with the LaunchDarkly MCP.

keelerm84 (approved):

PR #767, opened 2026-07-20
feat: Bound REPORT eval request body size with a 5 MiB default

keelerm84:
## Summary

Ports the REPORT eval request body size limit (#749, merged to v8) to v9, and -- because v9 is a major version -- ships a real **default limit of 5 MiB** that customers can override, rather than v8's unset/no-limit default.

### Background

`getClientSideContextProperties` read the entire REPORT request body into memory with an unbounded `io.ReadAll`. Combined with the intentional lack of an `http.Server` `ReadTimeout` (needed for streaming endpoints), an attacker could send an arbitrarily large body to the eval endpoints and drive the process to OOM (pen-test finding, SEC-8503).

### Changes

- New `[Main]` option `maxClientRequestBodySize` / `MAX_CLIENT_REQUEST_BODY_SIZE` (`ct.OptBase2Bytes`), with a `DefaultMaxClientRequestBodySize = 5 * units.MiB` constant.
- `getClientSideContextProperties` wraps the body read with `http.MaxBytesReader`; oversized bodies get `413 Request Entity Too Large`, other read errors get `400`.
- The limit is resolved once in `makeRouter` (config value or the 5 MiB default, always defined) and threaded through every eval/ping/poll REPORT handler -- covering the client-side (`/sdk/evalx/{envId}/context|user`), mobile (`/msdk/evalx/context|user`), server-side evalx REPORT, and the FDv2 poll/ping eval REPORT endpoints (the latter do not exist on v8).
- A non-positive explicit value (e.g. `0B`) is rejected at config load by `validateMaxClientRequestBodySize`, mirroring `validateMaxInboundPayloadSize`, so an invalid limit fails fast at startup.
- Docs updated to reflect the 5 MiB default; tests added for the `413` (limit configured) and no-limit (handler-level) paths plus the invalid-config case.

### Difference from v8 (#749)

On v8 the default is unset (no limit) for backwards compatibility. On v9 the default is `5MiB`, applied at the router so REPORT bodies are bounded by default; the handler still treats an undefined value as "no limit" so it can be disabled/overridden via config.

Relates to SEC-8503. Ports #749.

---

> [!NOTE]
> **Medium Risk**
> Changes request handling on all eval/ping REPORT paths with a new default cap that could reject unusually large legitimate contexts; mitigates prior unbounded memory read (SEC-8503).
> 
> **Overview**
> Adds **`maxClientRequestBodySize`** / **`MAX_CLIENT_REQUEST_BODY_SIZE`** on `[Main]` with a **5 MiB default** on v9, so REPORT/POST evaluation bodies are capped before context JSON is decoded.
> 
> **`getClientSideContextProperties`** now reads through **`http.MaxBytesReader`** when a limit is set; oversize bodies return **413** with a JSON error, and other read failures return **400**. **`makeRouter`** resolves the limit once (config or default) and passes it into client-side, mobile, server-side **evalx**, FDv2 **poll/eval**, and context-bearing ping stream handlers.
> 
> Startup validation rejects non-positive explicit values (e.g. **`0B`**), matching the inbound events payload check. Docs and tests cover the new option, **413** behavior, and the no-limit handler path when the opt is undefined.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 551776752dea1fa5c191e19eac47511af7dfebc5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


keelerm84:
@kparkinson-ld as discussed, the v9 port of this sets a default at 5MiB.

aaron-zeisler (approved):

PR #780, opened 2026-07-28
feat: Add init-concurrency limiter primitive and configuration

kinyoklion:
Adds the foundation for bounding **initialization-delivery concurrency** in the Relay Proxy: an admission-control limiter and a `[Concurrency]` configuration section. The limiter caps how many SDK *initialization deliveries* — the full-dataset payloads Relay materializes and sends when an SDK first receives its data — may be in flight at once, across both polling and streaming, so a reconnect storm can't drive unbounded resident-payload memory or egress.

This change adds **only** the limiter primitive and its configuration surface. Wiring it into the initialization paths is a follow-up; nothing consumes the limiter yet, so **behavior is unchanged** and the feature stays off unless configured.

## Limiter (`internal/concurrency`)
A two-limit keyed semaphore:
- `MaxConcurrent` — units of work held at once (in-flight deliveries).
- `MaxQueued` — bounded FIFO of waiters once concurrency is saturated; `0` sheds immediately.
- optional per-environment try-only gate, so one environment can't consume the whole budget and starve the others.
- `MaxConcurrent <= 0` ⇒ disabled (zero-overhead pass-through); nil-safe; `Acquire(ctx, envKey) (release, ok)`; `Close()` unblocks all waiters. Unit-tested under `-race`.

## Configuration (`[Concurrency]`)
| File key | Env var | Meaning |
|---|---|---|
| `maxConcurrent` | `INIT_MAX_CONCURRENT` | max in-flight deliveries; unset/`≤0` disables |
| `maxQueued` | `INIT_MAX_QUEUED` | bounded FIFO backlog; `0` = shed when full |
| `perEnvMaxPercent` | `INIT_PER_ENV_MAX_PERCENT` | one env's max share of the budget; `0` = global-only |
| `sendTimeout` | `INIT_SEND_TIMEOUT` | backstop to reclaim a slot from a stalled delivery (default 30s) |

`maxConcurrent` is an `OptInt` (not `OptIntGreaterThanZero`) so an explicit `0` disables gracefully instead of failing config validation.

## Testing
- Limiter unit tests under `-race`: admission, bounded-queue shed, disabled pass-through, exactly-once release, per-env gate.
- Config parse test for the `INIT_*` variables, including a regression that `INIT_MAX_CONCURRENT=0` disables rather than crashlooping.
- Existing `config` suite passes with the new section.

---

> [!NOTE]
> **Low Risk**
> New code is isolated; no production paths call the limiter yet, so default behavior is unchanged.
> 
> **Overview**
> Introduces a **foundation** for capping SDK initialization-delivery concurrency (polling and streaming). Runtime behavior is **unchanged** until a follow-up wires the limiter into init paths; with no config, the limiter stays disabled.
> 
> Adds `internal/concurrency` with an admission `Limiter`: global in-flight slots (`MaxConcurrent`), optional bounded wait queue (`MaxQueued`, `0` = reject when saturated), and an optional per-environment try-only gate so one env cannot take the whole budget. `Acquire(ctx, envKey)` returns a one-shot release; disabled/nil limiters admit with no overhead.
> 
> Adds `[Concurrency]` / `INIT_*` settings on `Config` (`INIT_MAX_CONCURRENT`, `INIT_MAX_QUEUED`, `INIT_PER_ENV_MAX_PERCENT`, `INIT_SEND_TIMEOUT`) and loads them from the environment. `MaxConcurrent` uses `OptInt` so `INIT_MAX_CONCURRENT=0` disables the feature without failing validation.
> 
> Unit tests cover limiter semantics (queueing, rejection, per-env isolation, context cancel, idempotent release) and config parsing including the zero-disable regression.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e3c6628bb6f80ca7d4e8dae7fb892bad2ba068cb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -141,8 +141,35 @@ type Config struct {
 	Filters     map[string]*FiltersConfig
 	Proxy       ProxyConfig
 	HTTP        HTTPConfig
+	Concurrency ConcurrencyConfig
 
 	OpenTelemetry OpenTelemetryConfig
+}
+
+// ConcurrencyConfig corresponds to the [Concurrency] section in the configuration file.
+//
+// It limits how many SDK initialization deliveries the Relay Proxy performs at the same
+// time, covering both polling and streaming. This caps the memory and egress that a burst
+// of connecting SDKs can consume. The limit is disabled unless MaxConcurrent is set.
+type ConcurrencyConfig struct {
+	// MaxConcurrent is the maximum number of initialization deliveries allowed in flight
+	// at once. A value of 0 or less disables the limit.
+	MaxConcurrent ct.OptInt `conf:"INIT_MAX_CONCURRENT"`
+
+	// MaxQueued is the maximum number of clients that may wait for a slot once
+	// MaxConcurrent is reached. A value of 0 rejects excess clients immediately instead
+	// of queueing them.
+	MaxQueued ct.OptInt `conf:"INIT_MAX_QUEUED"`
+
+	// PerEnvMaxPercent limits the share of the budget that any single environment may
+	// use, as a percentage of MaxConcurrent plus MaxQueued. This keeps one busy
+	// environment from starving the others. A value of 0 applies no per-environment
+	// limit.
+	PerEnvMaxPercent ct.OptInt `conf:"INIT_PER_ENV_MAX_PERCENT"`
+
+	// SendTimeout releases a delivery slot if a streaming initialization payload cannot
+	// make progress to its client within this duration. It defaults to 30s.
+	SendTimeout ct.OptDuration `conf:"INIT_SEND_TIMEOUT"`
 }
 
 // MainConfig contains global configuration options for Relay.
--- a/internal/concurrency/limiter.go
+++ b/internal/concurrency/limiter.go
@@ -1,0 +1,186 @@
+// Package concurrency provides an admission limiter that bounds how much concurrent
+// work a burst of requests or connections can impose on Relay. It uses two limits: a
+// maximum number of slots held at once, and a bounded FIFO queue of callers waiting for
+// a slot. An optional per-environment gate keeps one environment from using the whole
+// budget.
+package concurrency
+
+import (
+	"context"
+	"sync"
+	"sync/atomic"
+)
+
+// Params configures a Limiter. A MaxConcurrent of 0 or less produces a disabled limiter
+// whose Acquire always admits and does nothing.
+type Params struct {
+	// MaxConcurrent is the number of slots that may be held at once.
+	MaxConcurrent int
+	// MaxQueued is the number of callers that may wait in FIFO order for a slot once all
+	// slots are held. A value of 0 rejects callers immediately instead of waiting.
+	MaxQueued int
+	// PerEnvMax limits how many of a single environment's callers may participate,
+	// counting both held and waiting, at once. A value of 0 applies no per-environment
+	// limit. The per-environment gate never blocks; it only rejects.
+	PerEnvMax int
+}
+
+// Stats is a point-in-time snapshot of a Limiter's counters, for logging/metrics.
+type Stats struct {
+	Enabled       bool
+	MaxConcurrent int
+	MaxQueued     int
+	Held          int
+	Waiting       int
+	Admitted      int64
+	Rejected      int64
+}
+
+// Limiter bounds concurrency with two limits plus an optional per-environment gate that
+// never blocks and only rejects. The zero value is not usable; construct one with New.
+type Limiter struct {
+	name    string
+	enabled bool
+
+	tokens    chan struct{} // holds MaxConcurrent slots; receive to acquire, send to release
+	maxQueued int64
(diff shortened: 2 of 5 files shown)


config/config.go
```diff
@@ -141,10 +141,47 @@ type Config struct {
...
 	Proxy       ProxyConfig
 	HTTP        HTTPConfig
+	Concurrency ConcurrencyConfig
 
 	OpenTelemetry OpenTelemetryConfig
 }
 
+// ConcurrencyConfig corresponds to the [Concurrency] section in the configuration file.
+//
+// It bounds how many initialization deliveries -- the full-dataset payloads Relay
```
kinyoklion: Comment is too complex. Use simple sentences that says what it does and what it affects.

config/config.go
```diff
@@ -141,10 +141,47 @@ type Config struct {
...
+// materializes and writes to an SDK when it first receives its data -- may be in flight
+// at once, across polling and streaming for both FDv1 and FDv2. This bounds resident
+// payload memory (roughly MaxConcurrent times the dataset size) and egress concurrency
+// (each admitted client gets bandwidth / MaxConcurrent). Cheap operations -- FDv2
+// up-to-date replies, deltas, heartbeats, and pings -- are never gated.
+//
+// Every field defaults to disabled, so Relay behaves exactly as it does today unless an
```
kinyoklion: Comments shouldn't be relative to how something was. Every field defaulting to disabled means the concurrency limit is disabled. No need to reference "today".

config/config.go
```diff
@@ -141,10 +141,47 @@ type Config struct {
...
+// operator opts in.
+type ConcurrencyConfig struct {
+	// MaxConcurrent is the number of initialization deliveries that may be in flight at
```
kinyoklion: Again too long.

config/config.go
```diff
@@ -141,10 +141,47 @@ type Config struct {
...
+	// once. Unset, 0, or negative disables the limiter entirely (pass-through) -- this is
+	// deliberately an OptInt rather than OptIntGreaterThanZero so that an explicit 0 (a
+	// natural way to turn the feature off) disables gracefully instead of failing config
+	// validation and crashlooping the Relay.
+	MaxConcurrent ct.OptInt `conf:"INIT_MAX_CONCURRENT"`
+
+	// MaxQueued is how many clients may wait (bounded FIFO) when all slots are held.
+	// Waiters hold no payload buffer. 0 means never wait -- shed immediately when full.
```
kinyoklion: Anywhere we use a `--`, we should figure out something simpler.

internal/concurrency/limiter.go
```diff
@@ -0,0 +1,186 @@
+// Package concurrency provides an admission/concurrency limiter used to bound
+// how much work a request or connection herd can impose on Relay. It implements
+// a two-limit model (max concurrently-held tokens + a bounded FIFO wait backlog)
+// with an optional per-environment fairness gate, modeled on the streamer
```
kinyoklion: Don't mention streamer.

internal/concurrency/limiter.go
```diff
@@ -0,0 +1,186 @@
...
+	Rejected      int64
+}
+
+// Limiter bounds concurrency with two limits plus an optional per-environment gate that
+// never blocks and only rejects. The zero value is not usable; construct one with New.
+type Limiter struct {
+	name    string
+	enabled bool
+
+	tokens    chan struct{} // holds MaxConcurrent slots; receive to acquire, send to release
```
kinyoklion, line 45: Should we be using something like golang.org/x/sync/semaphore to support cancellation? Or does cancellation not matter much in our case? We would consume queue space, but be handled quickly once the channel was available.

kinyoklion: It appears the problem with the sync/semaphore would be the lack of queue limits.

keelerm84 (approved):

config/config.go
```diff
@@ -141,10 +141,37 @@ type Config struct {
...
+	// use, as a percentage of MaxConcurrent plus MaxQueued. This keeps one busy
+	// environment from starving the others. A value of 0 applies no per-environment
+	// limit.
+	PerEnvMaxPercent ct.OptInt `conf:"INIT_PER_ENV_MAX_PERCENT"`
+
+	// SendTimeout releases a delivery slot if a streaming initialization payload cannot
+	// make progress to its client within this duration. It defaults to 30s.
+	SendTimeout ct.OptDuration `conf:"INIT_SEND_TIMEOUT"`
+}
+
```
aaron-zeisler, line 174: Question: there's a pattern in the code for validating the Config object's elements: [[link](https://github.com/launchdarkly/ld-relay/blob/e3c6628bb6f80ca7d4e8dae7fb892bad2ba068cb/config/config_validation.go#L81)]. Would it be beneficial to perform that validation on these new parameters?

PR #781, opened 2026-07-29
feat: Add authenticated environment attributes to relay.auth span and metrics

keelerm84:
## Summary

Emit the authenticated environment as telemetry attributes so traces and metrics can be correlated by environment.

- **`relay.auth` span**: on successful auth, both auth middlewares now set `relay.auth.environment.name` (always, the human-readable display name) and `relay.auth.environment.id` (only when an `EnvironmentID` credential is configured). The `relay.auth.result=success` attribute remains unconditional.
- **Metrics**: the per-environment attribute set keeps the existing `environment.name` and now also emits `environment.id` when available.

The environment ID may be absent for SDK-key-only environments in a manual configuration, so it is only added when present.

---

> [!NOTE]
> **Low Risk**
> Observability-only attribute and middleware changes with broad test coverage; no auth or request-handling behavior changes beyond span/metric labels.
> 
> **Overview**
> Adds shared **`environment.name`** and **`environment.id`** telemetry so authenticated traffic can be joined across signals.
> 
> On successful auth, both server-side and client-side middleware call **`setEnvSpanAttributes`** on the **request span** (the otelmux parent), not on `relay.auth`, so every handler span in the trace inherits filterable environment context. Names use **`tracing.SanitizeAttributeValue`** (same rules as metrics); **`environment.id`** is omitted for SDK-key-only environments.
> 
> Metrics gain optional **`environment.id`**, reuse the same attribute keys from **`internal/tracing`**, and **`AddEnvironment`** now accepts an env ID. Per-environment attributes are immutable snapshots behind an atomic pointer; **`SetEnvironmentName`** (from **`SetIdentifiers`** on upstream renames) starts a new metric series under the new name while preserving env ID. Event metrics read attributes at record time so renames apply to async event counters too.
> 
> **`docs/metrics.md`** documents trace/metric correlation and rename behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 688a4918e9bf2433d82876e26b0f7e101adeca99. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/middleware/middleware.go
+++ b/internal/middleware/middleware.go
@@ -1,6 +1,7 @@
 package middleware
 
 import (
+	"context"
 	"encoding/base64"
 	"encoding/json"
 	"errors"
@@ -26,6 +27,7 @@ import (
 	"github.com/gorilla/mux"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
+	"go.opentelemetry.io/otel/trace"
 )
 
 const (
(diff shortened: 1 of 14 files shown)


internal/middleware/middleware.go
```diff
@@ -98,6 +99,20 @@ func parseApplicationTags(req *http.Request) (applicationID, applicationVersion
 	return
 }
 
+// authEnvSpanAttributes returns the relay.auth span attributes identifying the authenticated
+// environment: the human-readable environment name, plus the environment ID when an
+// EnvironmentID credential is configured (it may be absent for SDK-key-only environments in a
+// manual configuration).
+func authEnvSpanAttributes(clientCtx relayenv.EnvContext) []attribute.KeyValue {
+	attrs := []attribute.KeyValue{
+		tracing.AuthEnvNameKey.String(clientCtx.GetIdentifiers().GetDisplayName()),
```
kinyoklion: Display name needs sanitized.

keelerm84:
Context from a now-abandoned branch that may be useful to a reviewer: the attribute-snapshot rework here is load-bearing, not cosmetic.

While building an eventsource ServerTrace OTel bridge (#801, since closed unmerged), a multi-agent review found that `EnvironmentManager.GetAttributes()` on v9 is `attribute.NewSet(em.envKVs...)`, and `attribute.NewSet` sorts the slice it is given **in place** (otel v1.44.0 `attribute/set.go:238`, plus element swaps at `:256` that run even when keys are unique and already ordered). So that method is a write to `envKVs`, not a read of it.

On v9 today it is dormant only because `GetAttributes()` has zero production callers -- it is reachable only from `metrics_test.go`. #801 added a single innocuous-looking caller (`registerStreamChannel`, on the credential-rotation goroutine) and that was enough to produce a real data race against the request path, which reads `em.envKVs` unsynchronized via `buildRequestAttributes` / `buildDurationAttributes` on every request goroutine:

```
WARNING: DATA RACE
Write by goroutine 11: slices.SortStableFunc -> attribute.NewSetWithFiltered (set.go:238)
  -> metrics.(*EnvironmentManager).GetAttributes (metrics.go:297)
Previous read by goroutine 15: runtime.slicecopy
  -> metrics.buildRequestAttributes (constants.go:67)
```

Two notes in this PR's favour:

1. Returning a precomputed `set` from the immutable snapshot removes the sort-in-place write entirely, so a future caller cannot reintroduce the race by accident. That is the part that mattered for #801.
2. The capacity clip (`kvs[:len(kvs):len(kvs)]`) closes a second door that a simple defensive copy would leave open: an `append` to a slice with spare capacity writing into a slot another reader can see. Worth keeping, and worth keeping the comment that explains it.

Also worth noting the reorder is a genuine permutation rather than a same-value self-swap, since `envKVs` is built `[relay.id, environment.name]`, which is not in key order.

One caveat on the evidence: the race itself is proven under `-race`, but the downstream consequence (a torn multi-word `KeyValue` read producing a mislabeled point) is **not** -- 24k iterations produced zero corrupt sets, so the honest framing is undefined behavior the detector will flag, not "metrics are wrong today."

Finally: `internal/metrics/proof_rename_test.go` on this branch already covers the regression (`TestProofRaceRenameWhileRecording` drives `WithGauge`, `RecordRequestDuration`, `RecordEventsReceivedBytes`, the `EventMetricsRecorder` methods and `GetAttributes` itself, 4 goroutines each, against concurrent `SetEnvironmentName`). It passes under `-race` here. It is currently an untracked `proof_` file, so it would need committing under a normal name to actually guard this going forward.

PR #782, opened 2026-07-29
feat: Apply the init-concurrency limiter to polling and streaming

kinyoklion:
> [!IMPORTANT]
> In draft while its extracted pieces merge first — intended review order: #803 (limiter `Closed()`), #804 (initwrite `Cut()` + exported write slack), #805 (init-concurrency HTTP middleware), then #806 (budget construction from config; depends on #804). Their content is verbatim what this PR uses. Once they merge, this PR rebases down to the wiring: stream-provider integration, poll routes and endpoints, and docs.

Applies the init-concurrency limiter and the progress-aware write deadline (both already on this feature branch) to the endpoints that perform an *initialization delivery* — the full data-set payload the Relay Proxy serializes and sends when an SDK first receives its data. One shared budget covers polling and streaming, FDv1 and FDv2. This bounds the memory and egress that a burst of connecting or reconnecting SDKs can impose.

The limiter is disabled unless `[Concurrency]`/`INIT_*` is configured, so default behavior is unchanged. This PR is mostly wiring: the limiter primitive and the write-deadline writer merged separately, and the delta here is the middleware, the stream provider, the relay endpoints, and the documentation — plus one small addition to the writer (`Cut`, below) that the stream wiring needs.

## Coverage

Gated (full data set):
- Server-side streams `/all` and `/sdk/stream`. An FDv2 reconnect whose basis is current gets a small up-to-date reply, which is never charged.
- FDv2 polls `/sdk/poll` and `/sdk/poll/eval`, charged lazily on the full-basis branch only.
- FDv1 PHP all-flags poll `/sdk/flags`, charged on entry.

A queued request waits for a slot as long as its client is willing to. The client's disconnect is the designed exit: it relinquishes the queue spot immediately, and the SDK retries on its own backoff schedule. This is deliberate — SDKs already apply exponential backoff with jitter, so a relay-imposed wait bound would only move the retry earlier.

Not gated: single-item PHP lookups (`/sdk/flags/{key}`, `/sdk/segments/{key}` — per-evaluation traffic), `evalx` endpoints (evaluated results for one context), client-side ping streams (no payload), the legacy FDv1 flags-only stream, and deltas/heartbeats on established streams.

## Safety model

Giving up on a delivery must mean **closing the connection**, never leaving a partly-sent payload on an open one — a dropped connection is safe (the SDK discards the partial and reconnects), but a half-basis left on a live connection could be completed by a later delta into a corrupt data set. Everything below follows from that.

## Polling

A shed polling request gets **503 + jittered `Retry-After`**. While a slot is held, the response writer arms a progress-aware write deadline (throughput floor plus the `sendTimeout` cap); net/http clears the connection's deadline at end of request, so kept-alive connections are unaffected (pinned by a socket-level regression test).

## Streaming

The replay producer reads the store once (single-flight, keyed by basis), sends the free up-to-date reply when the client is current, and otherwise acquires a slot **before** serializing — so the budget bounds payload memory, and same-basis reconnects share one serialization. The slot is held **across the send**: the producer marks the delivery finished on every exit, triggers the handler's end-of-batch flush, and waits for the flushed basis or the end of the connection before releasing. `MaxConcurrent` therefore bounds concurrent sends and resident payloads, including the single-event FDv1 `/all` put.

A client that stalls without disconnecting is cut by the write deadline: a ~64 KB/s floor re-armed per write catches a stalled or too-slow reader quickly, and `sendTimeout` caps the whole delivery. The deadline is scoped to the delivery, so established streams (deltas, heartbeats, idle periods) carry none. A shed replay closes the connection (the SSE response has already started, so a 503 is not possible) and the SDK reconnects with backoff. A relay-initiated cut logs at warn, distinguished from an ordinary client disconnect.

When the client itself goes away mid-delivery, the slot frees at once — and a handler-scoped watcher cuts the blocked write in the same moment, so the payload's memory and egress end with the slot. Without the cut, a half-closing client's write would keep draining until its per-chunk deadline, and a reconnect loop of such clients could keep many payloads resident while holding no slots. The cut is a small addition to the writer, callable only within the handler's lifetime; a socket test pins the hostile shape end to end. All of this arms only when the limiter is enabled; a disabled deployment runs the exact base path.

## Validation

- Real-socket regression tests: slot held across the send, slot released after a healthy basis, stalled client cut at the deadline, busy stream surviving past the cap, deadline reaching the connection through the full middleware chain.
- Container repros (2 CPU / 2 GB, 3 MB data set): 250 stalled unique-basis FDv2 clients pin 840 MiB with the limiter disabled vs a flat 171 MiB at `maxConcurrent=16`; 120 stalled FDv1 clients at `maxConcurrent=1` hold flat ~37 MiB; 400 healthy churning clients are all served with no slot leak and a fresh client served immediately after.

## Follow-ups (tracked, not in this PR)

- Observability counters (deadline cuts vs cap cuts vs disconnects; limiter rejection causes) — planned metrics PR against the merged packages.
- Principal isolation: the budget is one shared pool by design; a per-trust-tier split can be layered on without primitive changes if needed.

---

> [!NOTE]
> **Overview**
> When **`[Concurrency]`** / **`INIT_*`** is set, the Relay Proxy applies one shared **initialization-delivery budget** to polling and server-side streaming so reconnect herds cannot pin unlimited memory and egress on full data-set payloads. With no config, behavior is unchanged.
> 
> **Relay wiring:** The server-side stream provider is created with **`WithInitLimiter`** and **`WithLogger`**. FDv1 **`/sdk/flags`** uses entry-time **`pollLimit`** (503 + **`Retry-After`** when shed). FDv2 **`/sdk/poll`** and **`/sdk/poll/eval`** use **`provideInitLimiter`** so handlers take a slot only on the full-basis branch; up-to-date replies stay free.
> 
> **Streaming:** **`withInitDeadline`** wraps SSE handlers with **`initwrite`** progress-aware write deadlines and a disconnect watcher that **`Cut()`**s blocked writes when the client leaves mid-delivery. Replay acquires the budget before serialization, re-reads the store after queue wait, sheds by closing the connection when the budget is full, and holds the slot through the full basis send via **`Begin`/`End`/`WaitAndFinish`**. **`peek`** is keyed by basis so concurrent reconnects at the current basis get up-to-date replies instead of stale full bases. **`sseLogger`** logs relay-initiated deadline cuts at **warn** vs ordinary disconnects at **debug**.
> 
> **Docs:** **`docs/configuration.md`** adds the **`[Concurrency]`** section (`maxConcurrent`, `maxQueued`, `sendTimeout`) and scope notes (gated vs excluded endpoints, shared pool, evalx/client-side init path).
> 
> **Tests:** Unit and real-socket tests cover shed/queue/shutdown paths, post-admission reads, slot lifecycle, watcher join, and stalled vs healthy clients.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8bc888a093ab2a16252bb5feb794bce947f68d25. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/middleware/concurrency.go
```diff
@@ -0,0 +1,45 @@
...
+// LimitConcurrency wraps a handler with a concurrency limiter keyed by the resolved
+// environment, for request-scoped initialization deliveries such as polling. It must be
+// applied inside an environment-selecting middleware stack so the EnvContext is present
+// on the request. The slot is held for the whole request (acquire on entry, release when
+// the handler returns) and acquired with the request context, so a client that
+// disconnects while queued drops out immediately. On shed it responds 503 with
+// Retry-After and does not invoke the wrapped handler. A disabled or nil limiter is a
+// pass-through with zero overhead.
+func LimitConcurrency(limiter *concurrency.Limiter) func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
```
keelerm84, line 58: We should add some metrics and spans for this.

internal/streams/stream_provider_server_side.go
```diff
@@ -203,115 +232,165 @@ func (r *serverSideEnvStreamRepository) replay(ctx context.Context, id string) c
...
+			r.sendEvents(ctx, out, MakeEventsForUpToDate(selector))
+			return
+		}
+
+		// This is a full-basis delivery. Draw from the shared budget before serializing,
+		// so the budget bounds both the memory of the payload we are about to build and
+		// the egress of sending it. The slot is held across the send and released when the
+		// send finishes, the client disconnects, or the stall backstop fires.
+		if r.initLimiter.Enabled() {
+			release, ok := r.initLimiter.Acquire(ctx, r.envKey)
```
keelerm84: Here too if we could manage it.

internal/middleware/concurrency.go
kinyoklion: Yes, I am currently deferring the otel instrumentation until we have the ServerTrace implementation in the eventsource. Then I will make a followup PR.

kinyoklion: Technically we only need it for half of the changes, but I think it is nice to put it in one batch.

config/config.go
```diff
@@ -161,12 +161,6 @@ type ConcurrencyConfig struct {
 	// of queueing them.
 	MaxQueued ct.OptInt `conf:"INIT_MAX_QUEUED"`
 
-	// PerEnvMaxPercent limits the share of the budget that any single environment may
```
kinyoklion: I've decided not to support this for now, so removing it from the config as well.

relay/relay_endpoints.go
```diff
@@ -230,6 +230,16 @@ func pollHandlerV2(w http.ResponseWriter, req *http.Request) {
...
 	} else {
+		// This is a full-basis delivery, whose serialization is the memory-heavy work the
+		// budget bounds. Draw a slot before building it (the cheap up-to-date branch above
+		// is never charged). The slot is held until the handler returns, covering the marshal
+		// and the response write.
+		release, ok := middleware.AcquireInitSlotFromContext(w, req)
+		if !ok {
+			return
+		}
+		defer release()
```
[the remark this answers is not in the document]
kinyoklion: This sounds fine that they will get the full payload sometimes on the next poll.

keelerm84 (approved):

PR #783, opened 2026-07-31
fix: End auth span before invoking the next handler

keelerm84:
The auth span used a deferred span.End() that fired only after the whole
middleware chain returned, so the span incorrectly measured all downstream
handling time. Wrap the auth logic in an inner scope so the span ends before
next.ServeHTTP is called.

---

> [!NOTE]
> **Low Risk**
> Observability-only structural change in auth middleware with regression tests; request auth outcomes and context values for handlers are intended to stay the same.
> 
> **Overview**
> Fixes **incorrect `relay.auth` span duration** in `SelectEnvironmentByAuthorizationKey` and `SelectEnvironmentByClientSideAuth`. Auth logic now runs in a short-lived inner scope so `defer span.End()` runs **before** `next.ServeHTTP`, so traces no longer attribute downstream handler time to authentication.
> 
> On success, environment and CORS data are merged onto the **original** request context (`downstreamCtx` from `req.Context()`), so the next handler is not left running under the ended auth span. Auth behavior and HTTP status responses are unchanged.
> 
> Adds middleware tests with an in-memory OTel span recorder to assert the auth span is finished when the next handler runs and that the active span in the next handler is not the auth span.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 44030becefd7226b7a8eb2ae2fb4d68254259c47. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/middleware/middleware.go
+++ b/internal/middleware/middleware.go
@@ -115,76 +115,91 @@ func Chain(middlewares ...mux.MiddlewareFunc) mux.MiddlewareFunc {
 func SelectEnvironmentByAuthorizationKey(sdkKind basictypes.SDKKind, envs RelayEnvironments) mux.MiddlewareFunc {
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
-			ctx, span := tracing.Tracer().Start(req.Context(), tracing.SpanAuth)
-			defer span.End()
-			span.SetAttributes(tracing.SDKKindKey.String(string(sdkKind)))
-			req = req.WithContext(ctx)
+			// Run authentication in its own scope so the span ends before we
+			// hand off to the next handler. Otherwise the deferred span.End()
+			// would fire only after the whole chain returns, and the auth span
+			// would incorrectly encompass all downstream handling time.
+			req, ok := func() (*http.Request, bool) {
+				// The request handed to the next handler must carry parentCtx, not the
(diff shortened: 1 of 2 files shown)


aaron-zeisler:
Is it possible to write a test to exercise this change?

internal/middleware/middleware.go
```diff
@@ -115,76 +115,91 @@ func Chain(middlewares ...mux.MiddlewareFunc) mux.MiddlewareFunc {
...
-				span.SetAttributes(tracing.AuthResultKey.String("invalid_credential"))
-				span.SetStatus(codes.Error, "invalid credential")
-				w.WriteHeader(http.StatusUnauthorized)
-				return
-			}
+			// Run authentication in its own scope so the span ends before we
+			// hand off to the next handler. Otherwise the deferred span.End()
+			// would fire only after the whole chain returns, and the auth span
+			// would incorrectly encompass all downstream handling time.
+			req, ok := func() (*http.Request, bool) {
```
kinyoklion: Do we need req out from this?

keelerm84: I think it just depends on whether or not you want to mutate it within this closure. I've updated the code to do that so you can compare the two approaches.

keelerm84:
> Is it possible to write a test to exercise this change?

Added some tests for this.

kinyoklion (approved):

PR #795, opened 2026-08-05
fix(concurrent-keys): reject SDK keys and mobile keys scoped to a view

aaron-zeisler:
## Summary

Makes the Relay Proxy reject SDK keys and mobile keys that are scoped to a view, so they never enter the accepted credential set. Filtering happens in `envfactory.BuildAcceptedSet` so a view-scoped key is absent from the credential lookup map entirely. An SDK presenting one gets a 401.

## Background

A view-scoped key (payload filtering v2) is only entitled to a subset of its environment's flags. Relay serves the entire environment payload and has no view support, so accepting one would silently over-deliver every flag in the environment to an SDK that only wants a subset.

## Commit details

The first two commits are the feature, the third is a cleanup task.

**1. `chore: carry the hasViews marker through the credential wire format`**

Adds a `hasViews` flag to each SDK key and mobile key the backend sends us, and passes it inward to where relay assembles an environment's credentials. Nothing acts on it yet.

**2. and 3. `fix: reject SDK keys and mobile keys scoped to a view`** 

Discards view-scoped keys while the payload is being read, so they never join the set of credentials an environment will accept and an SDK presenting one is turned away as unrecognized. An environment's own default SDK key and mobile key are deliberately exempt, since refusing either of those would take the whole environment offline.

**4. `refactor: make every credential-set rejection a malformed-payload error`**

When a payload left us with no usable SDK key, relay reported that as a different kind of error than it uses for other unusable payloads, which made offline mode log a misleading reason. They are now the same kind, so nothing downstream has to tell them apart.

## Integration coverage

Integration tests for this change are stacked in #796 .

## Open questions for review

**Should `/status` show the rejected view-scoped keys?**, so support can diagnose without asking the customer for logs? Deliberately excluded here: it would require carrying view-scoped values into `EnvContext`, reintroducing exactly the state the ingestion filter exists to avoid, and would add a field to a rep that was only just stabilized.

---

> [!NOTE]
> **Overview**
> Relay now reads **`hasViews`** on each `sdkKeys` / `mobileKeys` entry and **drops view-scoped credentials in `BuildAcceptedSet`**, so they never enter the accepted lookup map and SDKs using them get **401**. The environment **anchor SDK key** and **primary mobile key** are always kept even if marked view-scoped.
> 
> **`BuildAcceptedSet`** now returns **`(AcceptedSet, []string, error)`** — the string list is wire **key** identifiers that were filtered — and auto-config / offline handlers **WARN** when any were rejected. Stream **credential validation** still fails only on errors (malformed payloads), not on filtered keys.
> 
> An environment with **no usable SDK key** after filtering (empty `sdkKeys[]` or every entry view-scoped) is treated like other bad payloads via **`MalformedCredentialSetError`** (`no usable SDK key in sdkKeys[]`), replacing the separate builder error so offline mode logs consistently.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cef83932f283b43d4064373f217a139544e35964. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/autoconfig/stream_manager.go
+++ b/internal/autoconfig/stream_manager.go
@@ -501,8 +501,6 @@ func (s *StreamManager) dispatchEnvAction(id config.EnvironmentID, rep envfactor
 // set. It runs at the stream parse boundary, before the rep's version is recorded via Upsert: a
 // malformed payload must not advance the version, or the backend's fresh put — which carries the same
 // version — would be deduplicated away by the MessageReceiver.
-//
-// Not every error is a *MalformedCredentialSetError, so callers treat them all alike.
 func (s *StreamManager) validateCredentialPayload(rep envfactory.EnvironmentRep) error {
 	_, _, err := envfactory.BuildAcceptedSet(rep.ToParams())
 	return err
--- a/internal/autoconfig/stream_manager.go
+++ b/internal/autoconfig/stream_manager.go
@@ -498,16 +498,13 @@ func (s *StreamManager) dispatchEnvAction(id config.EnvironmentID, rep envfactor
 }
 
 // validateCredentialPayload checks that an environment rep carries a structurally valid credential
-// set. It is run at the stream parse boundary — before the rep's version is recorded via Upsert —
-// mirroring how an unparseable event is handled by gotMalformedEvent.
+// set. It runs at the stream parse boundary, before the rep's version is recorded via Upsert: a
+// malformed payload must not advance the version, or the backend's fresh put — which carries the same
+// version — would be deduplicated away by the MessageReceiver.
 //
-// A malformed credential payload must preserve the previous accepted set and force a
-// stream reconnect (RAC is one-way push with no NAK channel, so the reconnect is what makes the
-// backend resend a fresh put). Validating here rather than after Upsert is essential: the version is
-// not advanced, so the fresh put — which carries the same version — is not deduplicated away by the
-// MessageReceiver. Any error from BuildAcceptedSet is a *MalformedCredentialSetError.
+// Not every error is a *MalformedCredentialSetError, so callers treat them all alike.
 func (s *StreamManager) validateCredentialPayload(rep envfactory.EnvironmentRep) error {
-	_, err := envfactory.BuildAcceptedSet(rep.ToParams())
+	_, _, err := envfactory.BuildAcceptedSet(rep.ToParams())
 	return err
 }
 
--- a/internal/envfactory/env_params.go
+++ b/internal/envfactory/env_params.go
@@ -46,18 +46,22 @@ type EnvironmentParams struct {
 
 // AcceptedSDKKey is one entry in the accepted SDK key set for an environment.
 // Expiry is zero if the key is permanent.
+// HasViews is true if the SDK key is associated with a view.
 type AcceptedSDKKey struct {
-	Key    string
-	Value  config.SDKKey
-	Expiry time.Time
+	Key      string
+	Value    config.SDKKey
+	Expiry   time.Time
+	HasViews bool
 }
 
 // AcceptedMobileKey is one entry in the accepted mobile key set for an environment.
 // Expiry is zero if the key is permanent.
+// HasViews is true if the mobile key is associated with a view.
 type AcceptedMobileKey struct {
-	Key    string
-	Value  config.MobileKey
-	Expiry time.Time
+	Key      string
+	Value    config.MobileKey
+	Expiry   time.Time
+	HasViews bool
 }
 
 func (e EnvironmentParams) WithFilter(key config.FilterKey) EnvironmentParams {
--- a/internal/envfactory/env_rep.go
+++ b/internal/envfactory/env_rep.go
@@ -105,9 +105,10 @@ type ExpiringKeyRep struct {
 // Key is the human-readable identifier (non-secret, e.g. "default-sdk"); Value is
 // the credential secret (e.g. "sdk-xxxx-..."). See the EnvironmentRep TERMINOLOGY comment.
 type ConcurrentKeyRep struct {
-	Key    string `json:"key"`
-	Value  string `json:"value"`
-	Expiry *int64 `json:"expiry,omitempty"` // Unix-ms; nil = permanent
+	Key      string `json:"key"`
+	Value    string `json:"value"`
+	Expiry   *int64 `json:"expiry,omitempty"` // Unix-ms; nil = permanent
+	HasViews bool   `json:"hasViews"`
 }
 
 func ToTime(millisecondTime ldtime.UnixMillisecondTime) time.Time {
@@ -135,8 +136,9 @@ func (r EnvironmentRep) ToParams() EnvironmentParams {
 		params.AcceptedSDKKeys = make([]AcceptedSDKKey, 0, len(r.SDKKeys))
 		for _, k := range r.SDKKeys {
 			entry := AcceptedSDKKey{
-				Key:   k.Key,
-				Value: config.SDKKey(k.Value),
+				Key:      k.Key,
+				Value:    config.SDKKey(k.Value),
+				HasViews: k.HasViews,
 			}
 			if k.Expiry != nil {
 				entry.Expiry = time.UnixMilli(*k.Expiry)
@@ -162,8 +164,9 @@ func (r EnvironmentRep) ToParams() EnvironmentParams {
 		params.AcceptedMobileKeys = make([]AcceptedMobileKey, 0, len(r.MobileKeys))
 		for _, k := range r.MobileKeys {
 			entry := AcceptedMobileKey{
-				Key:   k.Key,
-				Value: config.MobileKey(k.Value),
+				Key:      k.Key,
+				Value:    config.MobileKey(k.Value),
+				HasViews: k.HasViews,
 			}
 			if k.Expiry != nil {
 				entry.Expiry = time.UnixMilli(*k.Expiry)
--- a/internal/envfactory/reconcile_helper.go
+++ b/internal/envfactory/reconcile_helper.go
@@ -1,30 +1,9 @@
 package envfactory
 
 import (
-	"github.com/launchdarkly/ld-relay/v8/config"
 	"github.com/launchdarkly/ld-relay/v8/internal/credential"
 	"github.com/launchdarkly/ld-relay/v8/internal/util"
 )
-
-// collectViewScopedValues returns the set of credential values that any entry marks as view-scoped,
-// excluding the designated key (the anchor, or the primary mobile key) which is never filtered.
-//
-// Two entries can carry the same value with only one of them marked. Keying on the value means one
-// marked entry rejects it, whichever position it holds.
-func collectViewScopedValues[E any, V comparable](entries []E, designated V, get func(E) (V, bool)) map[V]bool {
-	var viewScoped map[V]bool
-	for _, e := range entries {
-		value, hasViews := get(e)
-		if !hasViews || value == designated {
-			continue
-		}
-		if viewScoped == nil {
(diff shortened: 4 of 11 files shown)


internal/autoconfig/stream_manager.go
```diff
@@ -505,9 +505,17 @@ func (s *StreamManager) dispatchEnvAction(id config.EnvironmentID, rep envfactor
...
+// yields a plain error. Don't add an errors.As branch here without accounting for that.
+//
+// View-scoped keys are not a validity concern: they are filtered out of the accepted set, and the
+// handlers that actually apply the set log them. Reporting them here too would double-log every
+// payload, since this runs on all of them.
 func (s *StreamManager) validateCredentialPayload(rep envfactory.EnvironmentRep) error {
-	_, err := envfactory.BuildAcceptedSet(rep.ToParams())
+	_, _, err := envfactory.BuildAcceptedSet(rep.ToParams())
 	return err
 }
```
aaron-zeisler, line 507: Is this function necessary? It's a single line of code. I wonder if this function previously contained more validation code, but the validation code was removed or moved into BuildAcceptedSet during the concurrent keys project. Does that mean there's validation within BuildAcceptedSet that could/should be moved into this function?

internal/envfactory/env_params.go
```diff
@@ -46,18 +46,24 @@ type EnvironmentParams struct {
 
 // AcceptedSDKKey is one entry in the accepted SDK key set for an environment.
 // Expiry is zero if the key is permanent.
+//
+// HasViews mirrors ConcurrentKeyRep.HasViews. BuildAcceptedSet consumes it — dropping the entry, or
+// ignoring the marker when the entry is the anchor — so it never reaches the credential package.
```
aaron-zeisler: This comment should simply read "HasViews is true if the SDK key is associated with a view"

internal/envfactory/env_params.go
```diff
@@ -46,18 +46,24 @@ type EnvironmentParams struct {
...
+	Key      string
+	Value    config.SDKKey
+	Expiry   time.Time
+	HasViews bool
 }
 
 // AcceptedMobileKey is one entry in the accepted mobile key set for an environment.
-// Expiry is zero if the key is permanent.
+// Expiry is zero if the key is permanent. Mobile keys can be view-scoped too; the primary mobile key
+// is the designated key whose marker is ignored, as the anchor is for SDK keys.
```
aaron-zeisler: This comment should simply read "HasViews is true if the mobile key is associated with a view". The new comment should be written on the line below te existing comment about Expiry.

internal/envfactory/env_rep.go
```diff
@@ -108,6 +108,16 @@ type ConcurrentKeyRep struct {
...
 	Expiry *int64 `json:"expiry,omitempty"` // Unix-ms; nil = permanent
+
+	// HasViews is true when the key is scoped to one or more views, meaning it may only see a
+	// subset of the environment's flags. Relay serves the whole environment payload, so admitting
+	// such a key would silently over-deliver: BuildAcceptedSet keeps one out of the accepted set —
+	// except on the environment's default SDK key or default mobile key, where the marker is ignored
+	// rather than honored, because dropping a designated key would take the environment down.
+	//
+	// A plain bool, not *bool — an absent field means "not view-scoped", and there is no third
+	// state. (Expiry is a pointer only because nil there genuinely means "permanent".)
```
aaron-zeisler: This entire comment is unnecessary

internal/envfactory/reconcile_helper.go
```diff
@@ -1,10 +1,38 @@
...
+func collectViewScopedValues[E any, V comparable](entries []E, designated V, get func(E) (V, bool)) map[V]bool {
+	var tainted map[V]bool
+	for _, e := range entries {
+		value, hasViews := get(e)
+		if !hasViews || value == designated {
+			continue
+		}
+		if tainted == nil {
+			tainted = make(map[V]bool)
+		}
```
aaron-zeisler: Initialize the map where it's declared on line 22

internal/envfactory/reconcile_helper.go
```diff
@@ -1,10 +1,38 @@
...
+//
+// Keying on the value rather than testing each entry's own marker matters because the accepted-set
+// builder is first-wins: two entries can carry the same value, and if only the later one is marked,
+// testing entries individually would admit the credential via the earlier one while still reporting
+// it as rejected — a view-scoped credential silently accepted, with a log line claiming otherwise.
+// Tainting the value makes one marked entry sufficient to reject it, whichever position it holds.
+//
+// Excluding the designated key here rather than relying on case order is what keeps the anchor guard
+// structural: the returned set can never contain the anchor, so no ordering of the switch below can
+// drop it.
```
aaron-zeisler: Only keep the first paragraph of this comment.

internal/envfactory/reconcile_helper.go
```diff
@@ -27,9 +55,20 @@ import (
...
+//
+// Keys scoped to a view are filtered out here rather than at authentication time, making this the
+// single funnel for both RAC and the offline archive. The second return value names the keys that were
+// dropped, so callers can log what they lost; it holds the wire identifiers (the non-secret
+// human-readable names), never credential values. An SDK presenting one of them gets a 401 — the key is
+// simply absent from the lookup map. It is empty whenever an error is returned, since the caller
+// discards the whole payload.
+//
+// BuildAcceptedSet deliberately takes no logger: stream_manager.go calls it to validate every payload
+// before dispatch, and a logger there would double-log each one.
```
aaron-zeisler: Remove the 2nd paragraph from this comment.

The first paragraph of the comment should read:
```go
// Keys scoped to a view are filtered out here rather than at authentication time, making this the
// single funnel for both RAC and the offline archive. The second return value names the keys that were
// dropped, so callers can log what they lost. An SDK presenting one of them gets a 401: the key is
// simply absent from the lookup map.
```

relay/autoconfig_actions.go
```diff
@@ -11,8 +15,43 @@ const (
...
+// payload. Shared by the auto-config and offline-mode handlers, which are the only callers that apply
+// an accepted set (see validateCredentialPayload for why the parse-boundary caller stays silent).
+//
+// The identifiers logged are the wire "key" names, which are non-secret, so they appear unobscured —
+// they are what an operator needs to find the key in the LaunchDarkly UI.
+//
+// This logs once per payload that actually reaches a handler. Same-version payloads are dropped as a
+// no-op before dispatch (see autoconfig.MessageReceiver), so a plain reconnect does not re-log; a
+// version-advancing put or patch does. An environment with N payload filters logs N+1 times, once per
+// filtered copy, distinguishable by the filter key in the environment name.
```
aaron-zeisler: Reduce this comment to:

```go
// logViewScopedKeys reports the view-scoped credentials that BuildAcceptedSet filtered out of a
// payload. This logs once per payload that actually reaches a handler.
```

internal/autoconfig/stream_manager.go
```diff
@@ -498,16 +498,13 @@ func (s *StreamManager) dispatchEnvAction(id config.EnvironmentID, rep envfactor
...
+// set. It runs at the stream parse boundary, before the rep's version is recorded via Upsert: a
+// malformed payload must not advance the version, or the backend's fresh put — which carries the same
+// version — would be deduplicated away by the MessageReceiver.
 //
-// A malformed credential payload must preserve the previous accepted set and force a
-// stream reconnect (RAC is one-way push with no NAK channel, so the reconnect is what makes the
-// backend resend a fresh put). Validating here rather than after Upsert is essential: the version is
-// not advanced, so the fresh put — which carries the same version — is not deduplicated away by the
-// MessageReceiver. Any error from BuildAcceptedSet is a *MalformedCredentialSetError.
+// Not every error is a *MalformedCredentialSetError, so callers treat them all alike.
```
aaron-zeisler: I'm not sure what this error means. 1. Are all the errors actually MalformedCredentialSetError? If not, what are the other error types? And why is "so callers treat them all alike" specified? It seems like the errors should be handled differently if they are not all the same type.

internal/envfactory/env_rep.go
```diff
@@ -108,6 +108,8 @@ type ConcurrentKeyRep struct {
 	Key    string `json:"key"`
 	Value  string `json:"value"`
 	Expiry *int64 `json:"expiry,omitempty"` // Unix-ms; nil = permanent
+
+	HasViews bool `json:"hasViews"`
```
aaron-zeisler: Nit: no need for an empty line above this new property.

internal/envfactory/reconcile_helper.go
```diff
@@ -20,33 +41,47 @@ import (
...
+// non-empty params.AcceptedMobileKeys with no designated primary (params.MobileKey undefined), an
+// array entry with an empty value, or no usable SDK key at all. The caller must preserve the previous
+// accepted state and, for RAC handlers, reconnect the stream with jitter to force a fresh put. This is
+// the single home for the anchor invariant.
+//
+// Keys scoped to a view are filtered out here rather than at authentication time, making this the
+// single funnel for both RAC and the offline archive. The second return value names the keys that were
+// dropped, so callers can log what they lost. An SDK presenting one of them gets a 401: the key is
+// simply absent from the lookup map.
+func BuildAcceptedSet(params EnvironmentParams) (credential.AcceptedSet, []string, error) {
```
keelerm84, line 35: Seems like we return the rejected view-backed keys so we can log them. This is requiring us to iterate through the entries twice and maintain duplicate copies of the data.

Instead of doing all of that, can we not just iterate through the list a single time, and either log the rejection for each key as we encounter it, or even just a summary at the end?

keelerm84 (changes requested):

internal/envfactory/reconcile_helper.go
aaron-zeisler: Ah, thanks for pointing that out. Yeah, it seems like the code in BuildAcceptedSet could just perform one loop instead of two, completely trashing the helper method and its returned variable. I'll make that change.

In terms of logging, the callers of BuildAcceptedSet are currently logging that list. EnvironmentParams doesn't contain a logger. Is it OK if the callers perform the logging? Or is it common to define a global logger object within the code (that could be used ad-hoc within BuildAcceptedSet)?

aaron-zeisler: OK @keelerm84 I've made this change

internal/envfactory/reconcile_helper.go
```diff
@@ -79,7 +55,7 @@ func BuildAcceptedSet(params EnvironmentParams) (credential.AcceptedSet, []strin
 		case k.Value == anchor:
 			anchorInArray = true
 			b.WithAnchor(credential.SDKKeyParams{Value: k.Value, Key: util.PtrOrNil(k.Key)})
-		case viewScopedSDKValues[k.Value]:
+		case k.HasViews && k.Value != anchor:
```
keelerm84: This != anchor comparison isn't necessary. The earlier case would have caught this already.

internal/envfactory/reconcile_helper.go
```diff
@@ -109,7 +82,7 @@ func BuildAcceptedSet(params EnvironmentParams) (credential.AcceptedSet, []strin
 		case k.Value == params.MobileKey:
 			primaryMobileInArray = true
 			b.WithPrimaryMobileKey(credential.MobileKeyParams{Value: k.Value, Key: util.PtrOrNil(k.Key)})
-		case viewScopedMobileValues[k.Value]:
+		case k.HasViews && k.Value != params.MobileKey:
```
keelerm84: Same deal here.

keelerm84 (approved):

PR #801, opened 2026-08-07
feat: Add an OTel bridge for eventsource ServerTrace hooks

keelerm84:
## Summary

Adds the ld-relay side of the eventsource server observability work: an OTel bridge that fills in the `ServerTrace` callbacks released in eventsource v1.12.0, recording stream metrics and emitting child spans for SSE connections.

**Instruments** live on `metrics.Instruments`, created once in `NewManager` from the Manager's private meter, so they are noop when `USE_OTLP` is disabled. All under the `launchdarkly.relay.stream.*` prefix:

| Instrument | Type | Notes |
|---|---|---|
| `subscribers.active` | Int64UpDownCounter | |
| `connection.duration` | Float64Histogram (s) | attr `end.reason` |
| `subscribers.dropped` | Int64Counter | buffer overflow |
| `events.sent` | Int64Counter | attr `event.type`; live events only |
| `events.sent.size` | Int64Histogram (By) | live events only |
| `comments.sent` | Int64Counter | heartbeats |
| `events.discarded` | Int64Counter | attr `reason` (jitter coalescing) |
| `write.errors` | Int64Counter | |
| `replay.events` | Int64Histogram | events per drain |
| `replay.data.size` | Int64Histogram (By) | payload bytes per drain |
| `replay.drain.duration` | Float64Histogram (s) | |

Replayed events are encoded in bulk and flushed once per batch, so they do not fire `EventSent`. The `events.sent*` instruments therefore count individually-flushed live traffic only, and replay volume is reported through the `replay.*` instruments. A quiet environment showing no `events.sent` is expected.

**Bridge** (`internal/streams/otelbridge`) exposes `New(instruments, baseAttrs)` and `TraceFor(kind, protocol)`, which bakes `stream.kind` and `stream.protocol` into each returned trace. A concurrency-safe `RegisterChannel`/`UnregisterChannel` registry maps channel strings to per-environment attributes. Callbacks are non-blocking: one RLock-guarded map read plus an instrument recording. Unknown channels record against a `relay.id`-only fallback set so a measurement is never dropped. `event.type` is allowlisted to the types relay publishes, with anything else bucketed as `_other`.

**Spans.** When the request context carries a recording span (relay's otelmux middleware keeps one open per SSE connection), the bridge emits short child spans back-dated from the durations the library measured: `eventsource.write` per event/comment write and `eventsource.replay` per replay drain, plus `subscribed`, `unsubscribed`, and `write_error` span events on the request span. When no recording span is present, including the OTLP-disabled noop-tracer case, nothing is created.

**Replay truncation** is reported as two separate facts rather than one, because they answer different questions:

- `replay.aborted` is eventsource's own truncation verdict from `ReplayFinishedInfo.Aborted`.
- `replay.client_gone` reports whether the request context was already canceled when the drain ended.

These are deliberately not merged. `Aborted` only covers what eventsource observed, and relay's repositories end their batches early when the request context is canceled, which eventsource necessarily reports as a normal completion of whatever was produced. Overriding `Aborted` with the context state is no better, since a fully delivered batch whose client departs immediately afterwards would then land full-size samples in the aborted distributions. A drain truncated by cancellation surfaces as `aborted=false, client_gone=true`.

**Wiring.** Relay constructs the bridge only when `USE_OTLP` is enabled and passes its `TraceFor` into `NewStreamProvider`, which attaches a per-protocol `ServerTrace` to each of the 8 eventsource servers (4 stream kinds x FDv1/FDv2). When disabled, `Server.Trace` stays nil. Each environment registers its SSE channels with the bridge where it registers credentials with the SSE servers, and unregisters them on credential removal/rotation and environment teardown.

The first commit bumps eventsource to v1.12.0, which is where the `ServerTrace` callbacks landed. Every commit in this branch builds and passes the suite on its own.

Verified against a local relay exporting to an OTLP collector, across all four stream kinds: `server` (v1/v2), `server-flags` (v1), `mobile-ping` (v1/v2), and `js-ping` (v1).

## Known limitation

The `replay.aborted=true` and `replay.client_gone=true` paths are covered by unit tests but were not reproducible against a live relay. A replay batch small enough to fit in the kernel socket buffer drains in microseconds regardless of how slowly the client reads, so there is no mid-drain window to interrupt; for FDv1 `/all` the batch is a single `put` event, so no mid-batch state exists at all. Exercising these live needs a payload large enough to exceed the socket send buffer.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -21,7 +21,7 @@ require (
 	github.com/hashicorp/go-immutable-radix v1.3.1 // indirect
 	github.com/hashicorp/golang-lru v1.0.2 // indirect
 	github.com/kardianos/minwinsvc v1.0.2
-	github.com/launchdarkly/eventsource v1.11.2
+	github.com/launchdarkly/eventsource v1.12.0
 	github.com/launchdarkly/go-configtypes v1.2.2
 	github.com/launchdarkly/go-jsonstream/v3 v3.1.1
 	github.com/launchdarkly/go-sdk-common/v3 v3.5.0
--- a/go.sum
+++ b/go.sum
@@ -272,8 +272,8 @@ github.com/launchdarkly/api-client-go/v13 v13.0.1-0.20230420175109-f5469391a13e 
 github.com/launchdarkly/api-client-go/v13 v13.0.1-0.20230420175109-f5469391a13e/go.mod h1:cQRkOAs0LGcfIs6RSsHNqwhzItUZooyhpqPv0hgiQZM=
 github.com/launchdarkly/ccache v1.1.0 h1:voD1M+ZJXR3MREOKtBwgTF9hYHl1jg+vFKS/+VAkR2k=
 github.com/launchdarkly/ccache v1.1.0/go.mod h1:TlxzrlnzvYeXiLHmesMuvoZetu4Z97cV1SsdqqBJi1Q=
-github.com/launchdarkly/eventsource v1.11.2 h1:RwQcuYEQ7Gtz/f78AysiHpFLwDEHGRUduWs/+k+V2vs=
-github.com/launchdarkly/eventsource v1.11.2/go.mod h1:dU+rZxkPOlGPsyJPpiDqiepAcFwIITDUClY9+A6RrMw=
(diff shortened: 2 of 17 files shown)


keelerm84:
Abandoning this approach; closing unmerged. The eventsource v1.12.0 ServerTrace API this was built on is released and unaffected.

PR #802, opened 2026-08-11
feat: Count active requests for every endpoint, not just streams

keelerm84:
## Summary

The OTEL semantic convention defines `http.server.active_requests` as *"Number of active HTTP server requests"* (confirmed in `otel@v1.44.0/semconv/v1.26.0/metric.go`); its required attributes are `http.request.method` and `url.scheme`, and nothing in the definition is streaming-specific. Relay only incremented it inside the `CountServerConns` / `CountBrowserConns` / `CountMobileConns` / `CountClientConns` wrappers, which are applied to the SSE handlers only, so polling, event ingestion, goals, the gif pixel, status, and unmatched requests were all invisible.

The code and docs already disagreed about this: the instrument description in `internal/metrics/metrics.go` already read "Number of active HTTP server requests" while `docs/metrics.md` described "currently active stream connections from SDKs".

Active-request tracking moves out of the `CountXConns` wrappers into a `RequestMetrics` middleware. That middleware is the former `DurationMetrics`, renamed because it now does both; it already sat in all five middleware stacks after the environment selector, so every SDK endpoint is now counted with the correct `environment.name` and `platform.category`. Duration keeps its existing exclusion of streaming responses, whose lifetime is unbounded.

`CountXConns` is reduced to feeding only the stream-connection counts Relay reports back to LaunchDarkly. LaunchDarkly's notion of a "connection" is a held-open stream, so counting polls or event posts there would misreport usage. A test asserts these wrappers leave the active-requests instrument untouched, guarding against double counting.

Status endpoints and requests that matched no route have no environment, so they are tracked through a separate `UnscopedActiveRequests` middleware and report the `not-provided` sentinel for the environment and SDK attributes. For unmatched requests this is applied to the router's `NotFoundHandler` rather than registered with `Router.Use`, because gorilla/mux skips middleware entirely when a request matches no route (`mux.go:138-148`).

A new `relay.endpoint.type` attribute makes the broadened metric usable, with values `stream`, `poll`, `events`, `goals`, `status`, and `not-provided`. It is parameterized at each route registration site rather than inferred from the route template, so a route added later cannot be silently mislabeled. Two of the five stacks became functions of the endpoint type and `jsClientSideMiddlewareStack` gained a parameter; one structural change was needed, because `serverSideRouter.Use(serverSideMiddlewareStack)` covered both `/bulk` + `/diagnostic` (events) and `/all` + `/flags` (stream), so the stack is now applied per endpoint group instead of to the whole subrouter. Middleware ordering is preserved.

The attribute also lands on `http.server.request.duration` and `launchdarkly.relay.events.received.size`, which share the attribute builders. This adds no new series, since the endpoint type is functionally determined by `http.route`, which is already an attribute on both.

The increment and decrement share one pre-built attribute set; a mismatch between them would leak a permanently non-zero series.

### Notes for reviewers

- Filtering to `relay.endpoint.type=stream` reproduces the previous, streams-only meaning of this metric.
- Only streams stay open long enough to register as concurrent. Polls and event posts finish in microseconds, so their gauge sits at 0 between requests; the value is in the series existing at all, and in being able to break the metric down.
- A failed auth (401) is still not counted: the environment selector rejects the request before the metrics middleware runs.
- Relay never answers 405, so there is no method-mismatch gap. A wrong method falls through the catch-all `serverSideRouter` (`PathPrefix("")`) to the not-found handler and is counted as `not-provided`.

Closes SDK-2892

---

> [!NOTE]
> **Overview**
> Broadens `http.server.active_requests` to cover **every in-flight request** (poll, events, goals, status, unmatched), matching the OTEL semantic convention. Previously only SSE handlers incremented it.
> 
> Active-request tracking moves into renamed `RequestMetrics` middleware (formerly `DurationMetrics`). Stream wrappers (`CountXConns`) now only feed LaunchDarkly usage reporting, avoiding double-counting. A new `relay.endpoint.type` attribute (`stream`/`poll`/`events`/`goals`/`status`/`not-provided`) keeps the metric filterable; filtering to `stream` restores the old meaning.
> 
> Status and unmatched routes use `UnscopedActiveRequests` with `not-provided` for environment/SDK attributes. Also strips invalid UTF-8 from metric and span attributes so a hostile header cannot poison an entire OTLP export batch, and skips instrument/attribute construction when OpenTelemetry is disabled.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2339bdbdf49a8893ce32030a6692bb18b0a19f16. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/metrics/constants.go
+++ b/internal/metrics/constants.go
@@ -3,7 +3,8 @@ package metrics
 import (
 	"strings"
 	"time"
-	"unicode/utf8"
+
+	"github.com/launchdarkly/ld-relay/v9/internal/util"
 
 	"go.opentelemetry.io/otel/attribute"
 	semconv "go.opentelemetry.io/otel/semconv/v1.26.0"
@@ -124,18 +125,12 @@ func requestKVs(platform string, ri RequestInfo) []attribute.KeyValue {
 // Empty values are replaced with descriptive defaults, and slashes are replaced with underscores.
 // This is appropriate for user agent strings and SDK wrapper names, but not for routes.
 //
-// Invalid UTF-8 is stripped, which matters more than it looks. Attribute values end up in OTLP
-// protobuf string fields, and proto3 requires those to be valid UTF-8: a single bad byte fails the
-// marshal for the entire export batch, not just the offending data point. Because these series are
-// cumulative, the poisoned series is re-collected every interval, so exports keep failing until the
-// process restarts. Header values are not restricted to ASCII -- RFC 7230 permits obs-text, and Go's
-// HTTP parser passes those bytes through unchanged -- so any caller can supply one, including an
-// unauthenticated caller via the status and not-found handlers.
+// Invalid UTF-8 is stripped via util.SanitizeUTF8, which matters more than it looks: a single bad byte
+// fails the OTLP marshal for the entire export batch, and these series are cumulative, so exports keep
+// failing until the process restarts. Header values are not restricted to ASCII, so an unauthenticated
+// caller can supply one via the status and not-found handlers.
 func sanitizeTagValue(v string) string {
-	v = strings.ReplaceAll(v, "/", "_")
-	if !utf8.ValidString(v) {
-		v = strings.ToValidUTF8(v, "")
-	}
+	v = util.SanitizeUTF8(strings.ReplaceAll(v, "/", "_"))
 	if strings.TrimSpace(v) == "" {
 		return notProvidedValue
 	}
--- a/internal/metrics/constants.go
+++ b/internal/metrics/constants.go
@@ -3,6 +3,7 @@ package metrics
 import (
 	"strings"
 	"time"
+	"unicode/utf8"
 
 	"go.opentelemetry.io/otel/attribute"
 	semconv "go.opentelemetry.io/otel/semconv/v1.26.0"
@@ -122,11 +123,23 @@ func requestKVs(platform string, ri RequestInfo) []attribute.KeyValue {
 // sanitizeTagValue ensures attribute values are valid.
 // Empty values are replaced with descriptive defaults, and slashes are replaced with underscores.
 // This is appropriate for user agent strings and SDK wrapper names, but not for routes.
+//
+// Invalid UTF-8 is stripped, which matters more than it looks. Attribute values end up in OTLP
+// protobuf string fields, and proto3 requires those to be valid UTF-8: a single bad byte fails the
+// marshal for the entire export batch, not just the offending data point. Because these series are
+// cumulative, the poisoned series is re-collected every interval, so exports keep failing until the
+// process restarts. Header values are not restricted to ASCII -- RFC 7230 permits obs-text, and Go's
+// HTTP parser passes those bytes through unchanged -- so any caller can supply one, including an
+// unauthenticated caller via the status and not-found handlers.
 func sanitizeTagValue(v string) string {
+	v = strings.ReplaceAll(v, "/", "_")
+	if !utf8.ValidString(v) {
+		v = strings.ToValidUTF8(v, "")
+	}
 	if strings.TrimSpace(v) == "" {
 		return notProvidedValue
 	}
-	return strings.ReplaceAll(v, "/", "_")
+	return v
 }
 
 // sanitizeRouteValue ensures route attribute values are valid.
--- a/internal/metrics/constants.go
+++ b/internal/metrics/constants.go
@@ -21,9 +21,38 @@ const (
 
 	defaultFlushInterval = time.Minute
 
+	// notProvidedValue is the sentinel reported for an attribute whose value is absent.
+	notProvidedValue = "not-provided"
+
 	BrowserPlatformCategory = "browser"
 	MobilePlatformCategory  = "mobile"
 	ServerPlatformCategory  = "server"
+)
+
+// EndpointType identifies the kind of Relay endpoint that served a request. It is reported on
+// request-scoped metrics as the relay.endpoint.type attribute, so that a metric covering all
+// requests can still be broken down by the delivery mode the request belongs to.
+type EndpointType string
+
+const (
+	// EndpointTypeStream is an SSE endpoint that holds the connection open.
+	EndpointTypeStream EndpointType = "stream"
+
+	// EndpointTypePoll is a request/response flag delivery endpoint.
+	EndpointTypePoll EndpointType = "poll"
+
+	// EndpointTypeEvents is an analytics or diagnostic event ingestion endpoint.
+	EndpointTypeEvents EndpointType = "events"
+
+	// EndpointTypeStatus is a Relay status endpoint. These are not associated with an SDK or an
+	// LD environment.
+	EndpointTypeStatus EndpointType = "status"
+
+	// EndpointTypeGoals is the experimentation goals endpoint used by the JS SDK.
+	EndpointTypeGoals EndpointType = "goals"
+
+	// EndpointTypeNotProvided is used for requests that did not match any route Relay serves.
+	EndpointTypeNotProvided EndpointType = notProvidedValue
 )
 
 var (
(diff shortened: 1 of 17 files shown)


keelerm84:
Thanks — one of these held up and one didn't. Both are now answered by tests rather than prose.

## 405s: this router never reaches that branch

The reading of `mux.go:150-155` is right in the abstract — `MethodNotAllowedHandler` *is* assigned without building the middleware chain. But the branch is unreachable here, so wrapping it would be dead code.

The deciding detail is `route.go:69-79`: when a matcher **succeeds**, it clears a pending `ErrMethodMismatch`. For `POST /status`:

1. The `/status` route matches the path, fails the method matcher, sets `MatchErr = ErrMethodMismatch`.
2. Iteration reaches the catch-all `serverSideRouter` (`PathPrefix("")`), which matches every path. Its prefix matcher succeeds, so `ErrMethodMismatch` is cleared.
3. That subrouter has no route for `/status`, so it sets `ErrNotFound` — which `route.go:63-65` then clears too.
4. Top-level `Match` finishes with `MatchErr == nil`, skips the method-mismatch branch entirely, and lands on the `NotFoundHandler` branch, which this PR does wrap.

Confirmed against a running relay, not just read: `POST`/`PUT`/`DELETE`/`PATCH /status` all return **404**, and all four show up in `http.server.active_requests` under `relay.endpoint.type=not-provided`.

Worth noting the change wouldn't be merely redundant: if that branch *were* reachable, setting `MethodNotAllowedHandler` would turn those 404s into 405s, which is a user-visible API change.

`TestWrongMethodIsNotFoundAndIsCounted` now asserts 404-and-counted for four methods and documents the reasoning. If the catch-all subrouter is ever removed, it fails with a 405 — which is the signal to wrap `MethodNotAllowedHandler` at that point.

## Hot-path cost: confirmed, and fixed for the case that mattered most

First, a caveat on the suggested benchmark: `relay/relay_endpoints_benchmark_test.go` can't observe this change. Both it and the tracing benchmark use `buildPreRoutedRequest` and invoke handlers directly, never entering the middleware, so a before/after there would have shown nothing and been falsely reassuring.

So I added benchmarks that do enter it, comparing the shipped middleware against a reconstruction of the pre-change duration-only path. Per request, OTel **enabled**:

| | ns/op | B/op | allocs/op |
|---|---|---|---|
| duration-only (pre-change) | 5,477 | 5,262 | 23 |
| active+duration (this PR) | 8,367 | 8,764 | 32 |
| delta | +2,890 (+53%) | +3,502 (+67%) | +9 |

The cost is almost entirely attribute-set construction, not export: `buildRequestAttributes` alone is ~2,040 ns / 3,432 B. For scale, `BenchmarkEvaluateAllFlags` (50 flags, eval + serialize) is ~11,100 ns / 9,600 B, so a 50-flag poll grows roughly 16.4 us -> 19.2 us of measured work.

Chasing that turned up a **pre-existing** problem this PR was amplifying. With OpenTelemetry disabled, the Manager still built instruments from a noop meter; because they were non-nil, every recording path ran to completion — sanitizing, allocating and sorting attribute sets per request — before handing them to an instrument that throws them away. Operators exporting nothing were paying for telemetry that went nowhere, and this PR had doubled that.

`244eb342` leaves instruments nil when OpenTelemetry is disabled, so the existing nil checks short-circuit first. All ten recording sites were already guarded (a test now asserts each tolerates nil). OTel **disabled**:

| | ns/op | B/op | allocs/op |
|---|---|---|---|
| before this fix | 7,910 | 8,764 | 32 |
| after | 1,918 | 1,415 | 16 |

4.1x faster and 6.2x less allocation than before the fix, and 2.6x faster than the pre-active-requests code — so with OTel off the relay is now better off than before this PR, and active-request tracking costs essentially nothing (1,918 vs 1,862 ns against duration-only, within noise).

The +53% on the **enabled** path is real and remains. Options I looked at and did not take, happy to pursue any of them: memoizing attribute sets keyed on the sanitized tuple (biggest win, but the key includes `instance.id` and `application.version`, so it needs a bounded cache); or trimming attributes on `active_requests` specifically, since semconv only requires `http.request.method` and `url.scheme` (cheaper, but loses the user-agent / application / instance breakdowns on that instrument). Say the word if you'd rather one of those landed here instead of as a follow-up.

Re-verified end to end against a live relay after the perf change: 50 series, 44 route templates, all six endpoint types, every route matching its expected type, zero non-zero series once streams closed.

internal/metrics/constants.go
```diff
@@ -47,58 +77,54 @@ var (
...
 // sanitizeTagValue ensures attribute values are valid.
 // Empty values are replaced with descriptive defaults, and slashes are replaced with underscores.
 // This is appropriate for user agent strings and SDK wrapper names, but not for routes.
 func sanitizeTagValue(v string) string {
 	if strings.TrimSpace(v) == "" {
-		return "not-provided"
+		return notProvidedValue
 	}
 	return strings.ReplaceAll(v, "/", "_")
 }
```
aaron-zeisler, line 138: My agent is pointing out an issue in this function: utf-8 is expected, but the function doesn't sanitize or validate for utf-8. It's suggesting this fix:
```go.
if !utf8.ValidString(v) {
	v = strings.ToValidUTF8(v, "")
}
```
(`notProvidedValue` or another constant may be better than the empty space in the code example)

keelerm84: Good catch, and it turned out to matter more than "expected UTF-8" suggests. I chased each link rather than assuming, because my first instinct was that header values would be ASCII-constrained anyway. They are not.

**Invalid UTF-8 is reachable.** RFC 7230 permits obs-text (0x80-0xFF) and Go's HTTP parser passes those bytes to the handler unchanged, so `User-Agent: bad-\xff\xfe-agent` arrives intact.

**It fails the marshal, and not just for the bad point.** `proto.Marshal` returns `string field contains invalid UTF-8`. I recorded one clean series and one poisoned series on the same instrument and flushed to a test collector: `ForceFlush` errored with **0 export requests and 0 bytes delivered**. The clean series never left the process.

**For metrics it is unrecoverable.** These series are cumulative, so the poisoned one is re-collected every interval. A second flush, after recording only clean data, also exported nothing. Metrics stay dead until the process restarts.

**And on this PR it became reachable without credentials.** The metrics middleware sits behind auth, but `UnscopedActiveRequests` -- which this PR adds for `/status` and unmatched routes -- reads `User-Agent` and requires none. Confirmed end to end: unauthenticated `GET /no/such/route` and `GET /status` with a malformed User-Agent both produced poisoned series. So `curl -H 'User-Agent: <invalid bytes>' http://relay/anything` permanently disables all metrics export.

The underlying gap predates this branch, since `user_agent` and the application tags were already attributes on the duration histogram. What is new is that it no longer takes a valid SDK key.

Fixed in `f9877e03`, structured as you suggested, with two adjustments: the strip runs after the slash replacement, and a value that is entirely invalid collapses to the `not-provided` sentinel rather than an empty string.

While confirming the blast radius I checked whether spans have the same exposure, since they use the same protobuf. They do, via three attributes -- `user_agent.original` and `client.address` (the latter from `X-Forwarded-For`, used unvalidated by the instrumentation) and `url.path` from the decoded request path -- plus our own `relay.store.key`, which is a decoded path variable. Traces recover on the next batch rather than permanently, but one hostile request still costs every span batched with it. `2339bdbd` covers those.

Both fixes are verified by tests that assert the invariant broadly rather than checking the helper: every string attribute recorded for a request with hostile headers, across the unauthenticated paths too. I confirmed they are not vacuous by reverting each fix -- 15 failures on the metrics side, and the span test named `user_agent.original` and `client.address` specifically.

One limitation worth flagging: the span fix is a fixed list of three attributes, because the instrumentation sets them and `ReadOnlySpan` cannot be rewritten at the export boundary. If a contrib upgrade starts recording another request-derived attribute, that list goes stale silently -- so `TestSpanAttributesAreValidUTF8` checks every string attribute on every span, and will fail and name the new one if that happens.

kinyoklion (approved):

PR #817, opened 2026-08-14
feat: Support concurrent SDK keys and mobile keys

aaron-zeisler:
## Summary

Merges the `feat/concurrent-keys` feature branch into `v8`, bringing concurrent SDK key and mobile key support to the Relay Proxy. 58 commits, 78 files changed.

## Background

Relay previously modelled an environment's credentials as a single SDK key plus an optional expiring key, which meant it could only ever authenticate one active key at a time and had to tear down and rebuild its upstream SDK client whenever that key changed. Concurrent keys generalizes this: an environment carries a set of accepted SDK keys and mobile keys delivered by Relay Auto Config or the offline archive, any of which authenticates a downstream connection, while a single designated *anchor* key owns the upstream SDK client. Rotation becomes a re-anchor — the upstream client, data store, and big-segment sync hand over to the new anchor without dropping connected clients.

## Changes

- **Credential model** — `Rotator` now tracks an accepted key set with wire identifiers instead of a single key plus grace-period key; the legacy single-key rotation path is gone.
- **Re-anchoring** — the SDK anchor flip is deferred to an explicit `CommitAnchor`, performed synchronously, with data store handover, big-segment sync re-anchoring, and closure of the demoted anchor's client.
- **Key plumbing** — `sdkKeys[]` and `mobileKeys[]` are carried through RAC and the offline archive; payloads whose primary mobile key is absent from `mobileKeys[]` are rejected, as are keys scoped to a view.
- **Status endpoint** — `/status` surfaces the full `sdkKeys[]` and `mobileKeys[]` arrays.
- **Stream handlers** — built on demand rather than one per credential.
- **Tests** — integration coverage for multi-key downstream auth, re-anchor behavior and rollback, lifecycle, and view-scoped key rejection.

## Merge notes

There are no `v8` commits missing from this branch. Every functional commit that landed on `v8` during development has been brought over — six re-landed as their own PRs (#708, #711, #727, #740, #769, #771) and four via the catch-up in #787 (#734, #749, #750, #789). All of `go.mod`, `go.sum`, `config/`, `docs/`, `internal/browser/cors.go`, `internal/events/event_publisher.go`, `relay/relay_endpoints.go`, and `relay/relay_routes.go` are byte-identical to `v8`.

`relay/version/version.go` was the only conflicting file and is now aligned with `v8` at `8.20.0`, so `.release-please-manifest.json` and the version string stay consistent on the production branch. The `8.21.0-rc.1` string it previously held is preserved by the tag of the same name. Since `v8` already shipped 8.20.0 without concurrent keys, the GA release off this merge needs to be 8.21.0 or later.

[SDK-2535]: https://launchdarkly.atlassian.net/browse/SDK-2535?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Overview**
> Relay can now authenticate **multiple concurrent SDK and mobile keys** per environment instead of one primary plus an optional expiring SDK key. AutoConfig and the offline archive carry `sdkKeys[]` / `mobileKeys[]`; any accepted key authenticates downstream traffic, while a single **anchor** owns the upstream SDK client.
> 
> **Rotation is a re-anchor**, not a tear-down of the environment. `Rotator.Reconcile` diffs the full accepted set (including per-key expiry and mobile keys). The env context adds new mappings first, then `CommitAnchor` after the new client is ready (store handover, big-segment resync, close the old client). Failed re-anchors revert only the anchor change. Malformed RAC credential payloads are rejected **before** version upsert, the stream reconnects, and the AutoConfig cache keeps last-good entries.
> 
> `/status` now returns full `sdkKeys` / `mobileKeys` arrays (obscured values, optional names/expiry). Stream handlers are built on demand from the accepted set. View-scoped keys are dropped (401). Analytics still go upstream under the **anchor**; diagnostics still proxy the incoming Authorization header.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 621a0411ee6a06d3f890c6a631d9362068be6557. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/relayenv/env_context_impl.go
```diff
@@ -552,16 +537,299 @@ func (c *envContextImpl) SetIdentifiers(ei EnvIdentifiers) {
...
+
+	// The new anchor's client is now authoritative, so close the previous anchor's client. The shared
+	// store wrapper survives: it is refcounted and the new anchor's client holds it. Offline mode is
+	// exempt because it built no replacement, and its single file-data client must keep serving.
+	if !c.offline {
+		if oldClient := c.clients[previousAnchor]; oldClient != nil {
+			delete(c.clients, previousAnchor)
+			_ = oldClient.Close()
+		}
+	}
```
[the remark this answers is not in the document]
aaron-zeisler: I had my local agent dig into this and it's a false positive (Claude wrote tests to try to exercise the issue and couldn't reproduce it).

internal/credential/rotator.go
```diff
@@ -60,9 +83,11 @@ func (r *Rotator) Initialize(credentials []SDKCredential) {
 		}
 		switch cred := cred.(type) {
 		case config.SDKKey:
-			r.primarySdkKey = cred
+			r.anchorKey = cred
```
keelerm84, line 86: This bit seems a little confusing. Last credential provided "wins" anchor position. However, the only call site provides a single SDK key.

Should we change the signature to just explicitly accept one of each type and then we can remove this loop in favor of direct comparisons?

keelerm84 (approved):

internal/credential/rotator.go
aaron-zeisler: Yeah, this function's signature didn't change, but the number of possible SDK keys in the system did. I think it would make sense to change the signature to be more explicit about that. I'll open a pull request but it'll be after the initial release.

PR #836, opened 2026-08-19
fix(credential): keep the old SDK key valid until the new key takes over

aaron-zeisler:
## Summary

Relay serves each environment through one SDK key. That key is the anchor.

A rotation replaces the anchor. Relay builds a client for the new key first. The anchor moves only after that build finishes. The build takes up to `Main.InitTimeout`, which is 10 seconds by default.

When a payload revoked the old key, `Reconcile` dropped it from the accepted set at once. Relay kept serving on that key for the whole build. So Relay served on a key it no longer accepted.

`GetStreamHandler` checks the accepted set. During the build it answered 404 for the key Relay was serving on. Any SDK that connected or reconnected in that window got that 404.

This PR keeps the old key accepted until the anchor moves.

## Background

A 404 from a stream endpoint is terminal for an SDK data source. Only 400, 408 and 429 are recoverable. An SDK that reconnects inside the window can stop its data source for good.

The window also broke a documented rule for `/status`: the `sdkKeys[]` array always contains the anchor. During the build it did not.

This defect is older than the re-anchor retry. An ordinary successful rotation opens the same window. The retry reopens the window once per credential cleanup interval while a build keeps failing. That is how it surfaced.

The rotator already applies this rule in two places. `StepTime` never expires the anchor, and `DeprecatedCredentials` excludes it. This PR extends the same rule to `Reconcile`.

## Changes

`reconcileSDKKeys` restores the anchor's entry when the diff drops it, so the key keeps its expiry and its name. `CommitAnchor` deletes that entry once the anchor moves, and the revocation queued during reconcile still removes the key's mappings straight after.

### DIFF
--- a/internal/credential/rotator.go
+++ b/internal/credential/rotator.go
@@ -33,6 +33,10 @@ type AcceptedKeySet struct {
 type Rotator struct {
 	loggers ldlog.Loggers
 
+	// anchorPendingRemoval means a payload revoked the anchor, but it stays accepted because it is
+	// still serving. CommitAnchor drops it when the anchor moves. See reconcileSDKKeys.
+	anchorPendingRemoval bool
+
 	// There is only one mobile key active at a given time.
 	primaryMobileKey config.MobileKey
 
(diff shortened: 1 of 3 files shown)


aaron-zeisler:
Superseded: this fix is now commit d1b8d4fb directly on `feat/concurrent-keys`, so it is reviewed as part of #817.

PR #876, opened 2026-09-18
feat: Export the status document as OpenTelemetry metrics

keelerm84:
## Summary

Relay's health and per-environment state are readable only by polling `/status` and parsing the nested JSON. That leaves no way to alert on a transition, and no way to ask a fleet-wide question at all, since the document is per process. This reports the same state through the OTel pipeline Relay already has.

The snapshot is built by the code that serves the endpoint, so a metric cannot describe a state the document would not. The cost is that a collection performs the same big segment store query the endpoint does.

**Encoding.** A state field is reported as one series per possible state, with the state on `launchdarkly.relay.state` and the current state reading 1. The zeros are emitted deliberately: a backend that keeps answering with the last sample of a series that stopped being written would otherwise show two states at 1 across a transition. The instruments carry annotation units rather than the dimensionless `1`, which Prometheus renames with a `_ratio` suffix.

**Tiers.** The relay-level metrics, including the fleet counts by state and the auto-configuration stream, are exported whenever OTLP export is enabled; their cardinality is fixed at about twenty series. The per-environment series are off by default behind `environmentStatusMetrics` / `OTEL_ENVIRONMENT_STATUS_METRICS`, because their cardinality follows the environment count rather than the traffic — in automatic configuration mode that is every environment in the account, idle ones included, and the OTel cardinality cap discards series silently once it is reached.

**Also here.** The status vocabulary (`healthy`, `degraded`, `connected`, `disconnected`, and the data store states) moves into `internal/api` as constants so the endpoint and the metrics cannot drift. The relay version and the embedded Go SDK version become resource attributes, since neither varies within a process. Two pre-existing documentation errors are fixed: `OTEL_EXPORTER_OTLP_ENDPOINT` was documented with a `/v1/metrics` path the SDK appends itself (our own playbooks configuration uses the base URL), and the Prometheus naming notes omitted the `_seconds` suffix.

`docs/metrics.md` also records three things these metrics cannot tell you, each of which would otherwise produce a wrong alert: the data store state is not a store liveness check, a stopped relay keeps reading healthy for the length of the backend's lookback, and a `last_error` series is superseded rather than replaced when the kind of failure changes.

---

> [!NOTE]
> **Overview**
> Adds **OpenTelemetry status metrics** that mirror `/status`, built from the same snapshot path (`collectEnvironmentStatuses` / `statusSnapshot`) so metrics and the JSON document cannot diverge. Relay-level gauges (health, environment counts by state, auto-config stream) export whenever OTLP is on; **per-environment series stay off by default** behind `environmentStatusMetrics` / `OTEL_ENVIRONMENT_STATUS_METRICS` because cardinality scales with environment count (~18 series each).
> 
> State fields use **one 0/1 series per possible value** on `launchdarkly.relay.state`; optional per-env metrics use `launchdarkly.environment.name` (display name) for joins with request metrics. Shared status strings move to **`internal/api` constants**; **`service.version`** and **`launchdarkly.relay.sdk.version`** are added as OTEL resource attributes.
> 
> Docs expand `metrics.md` (query/alert guidance, cardinality caveats, limitations) and fix Prometheus **`OTEL_EXPORTER_OTLP_ENDPOINT`** (base URL, not `/v1/metrics`). Config validation tests cover the new boolean (including explicit `false`/`0`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit edbe77f9a75159423af48344b0f534554bee4506. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/metrics/status_measures.go
```diff
@@ -0,0 +1,442 @@
...
+	// environmentUnit is the unit for a gauge that counts environments, or that reports one series
+	// per environment. It is also an annotation unit, so it adds no suffix either.
+	environmentUnit = "{environment}"
+
+	// timestampUnit is seconds, which Prometheus renders as a _seconds suffix.
+	timestampUnit = "s"
+)
+
+// The states each field can report. Every one is observed at each collection, including the zeros.
+// Observing only the current state would leave the others without a fresh sample, and a backend
```
kinyoklion: This is a confusing assemblage of words.

It is trying to convey something like these being instantaneous samples?

kinyoklion (approved):

internal/metrics/status_measures.go
keelerm84: Reworded both. The old comment tried to explain the zeros in one sentence and landed on something that read like instantaneous samples. It now says it plainly: Relay observes every state each collection, a series that stops being observed keeps its last value in the backend for five minutes, so dropping the zeros would report two states as current at once. `observeState` had the same shape and got the same treatment.

PR #887, opened 2026-09-21
feat: Allow disabling OTel log, trace and metric export individually

keelerm84:
USE_OTLP was a single all-or-nothing switch: turning it on started log,
trace and metric export together, and there was no way to run two of the
three. Operators whose collector cannot accept one signal had to give up
all of them.

That is the common case rather than an edge case. A Datadog Agent accepts
OTLP traces and metrics out of the box but leaves log ingestion disabled,
so pointing Relay at a default-configured Agent means the log export
fails on every attempt. On a host deployment the export is redundant
anyway once something already collects Relay's console output.

Honor the OpenTelemetry-specified OTEL_LOGS_EXPORTER,
OTEL_TRACES_EXPORTER and OTEL_METRICS_EXPORTER, where "none" disables
that signal. USE_OTLP remains the master switch, and an unset per-signal
setting still exports, so existing configurations behave exactly as
before.

The Go SDK ships no autoconfiguration package, so Relay reads these
itself. They are declared under their specification names rather than as
a Relay-specific flag: operators arriving from another LaunchDarkly SDK
reach for the variable that already works, and mirroring them into
OpenTelemetryConfig the way OTEL_EXPORTER_OTLP_PROTOCOL already is means
they can also be set in the configuration file. Relay implements only the
OTLP exporter, so it accepts "otlp" and "none" and rejects anything else
at startup instead of ignoring a value the operator believed had taken
effect.

Also corrects the documentation, which described USE_OTLP as enabling
metric export only. Log and trace export were undocumented, so nothing
told an operator to expect either.

---

> [!NOTE]
> **Overview**
> Adds **per-signal OTLP export control** via the standard `OTEL_LOGS_EXPORTER`, `OTEL_TRACES_EXPORTER`, and `OTEL_METRICS_EXPORTER` settings (env or `[OpenTelemetry]` file section). **`USE_OTLP` stays the master switch**; unset or `otlp` still exports each signal, and **`none` turns off that signal only**.
> 
> Startup and the metrics manager now gate work on **`ExportLogs()` / `ExportTraces()` / `ExportMetrics()`** instead of a single `Enabled` check—so e.g. `OTEL_METRICS_EXPORTER=none` skips building the meter provider and instruments, matching full OTLP-off behavior.
> 
> **Unrecognized exporter values** (`zipkin`, `prometheus`, `console`, comma lists) **log a warning and are ignored**, leaving OTLP export on so pod-wide OTEL env vars do not block Relay from starting.
> 
> Docs are updated to describe logs/traces alongside metrics and to document Datadog’s separate OTLP log switch vs `OTEL_LOGS_EXPORTER=none`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ca82b48808610d6c436e62bfbabcbaaa598eec9b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


config/config.go
```diff
@@ -351,8 +352,48 @@ type HTTPConfig struct {
 // environment variable for the cardinality limit, and the Go SDK's own OTEL_GO_X_CARDINALITY_LIMIT sits
 // in its experimental namespace, so Relay owns this setting. When it is undefined, Relay applies no
 // option and the SDK's default (or OTEL_GO_X_CARDINALITY_LIMIT, if the operator set it) stands.
+//
+// LogsExporter, TracesExporter and MetricsExporter are read by Relay rather than by the SDK for a
+// different reason: the specification defines them, but only the autoconfiguration packages that other
+// languages ship implement them, and the Go SDK has none. They are declared here, under their
+// specification names, so that operators arriving from another LaunchDarkly SDK reach for the variable
+// that already works and so that the setting can also be given in the configuration file. Relay only
```
kinyoklion: This seems a little scary as these could be set for a host/pod/etc and they could be set to console or zipkin. Those pre-existing settings could then impact relays ability to start instead of just its ability to export logs.

kinyoklion (changes requested):

config/config.go
keelerm84: The configuration-types guidance says that for an enum-valued variable, when the implementation doesn't recognize the value it "MUST generate a warning and gracefully ignore the setting."

Fixed in 517b3b26: unrecognized values warn and are ignored, falling back to the spec's `otlp` default — Relay's behavior before it read these variables. Only `none` turns a signal off.

One thing worth your explicit agreement, since the warning doesn't remove it: with the spec names, an inherited `OTEL_TRACES_EXPORTER=none` set pod-wide *will* quietly disable Relay's trace export. That's what every other OTel SDK does with that variable, so I think it's correct, but it is the other direction of the same ambient-config exposure you flagged. The alternative is Relay-specific names, which costs the cross-SDK familiarity that motivated using the spec names.

kinyoklion (approved):

